PR #7, opened 2018-03-28
Sync artifacts dir fix

apostergiou:


apostergiou:
The tests are failing without the `"/*"` postfix because the end-to-end test seems to always create the `artifacts` parent directory.

I am marking this as **wip** till I found out what causes this behavior.

PR #12, opened 2018-04-03
Clean all pending builds on mistry boot time

apostergiou:
This fixes the following bug:

- a job {"project": "hanging-pending"} hasn't finished building and the service shuts down
- when the service restarts the build is still pending but hasn't resumed building, it will stay pending forever

The result is that the client is not able to fetch the jobs results or request another one (with the same parameters) since the service _sees_ the build as pending.

We solve this by pruning all pending builds on the service boot time.

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,8 +16,10 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/skroutz/mistry/btrfs"
@@ -99,6 +101,12 @@ func main() {
 			return fmt.Errorf("invalid filesystem argument (%v)", fsList)
 		}
 		curfs = fs
+
+		err = PruneZombieBuilds(curfs)
+		if err != nil {
+			return err
+		}
+
 		return nil
 	}
 	app.Action = func(c *cli.Context) error {
@@ -122,4 +130,27 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+}
+
+func PruneZombieBuilds(curfs FileSystem) error {
+	projects, err := ioutil.ReadDir(cfg.ProjectsPath)
+	if err != nil {
+		return err
+	}
+	l := log.New(os.Stderr, "[cleanup] ", log.LstdFlags)
+
+	for _, p := range projects {
+		pendingPath := filepath.Join(cfg.BuildPath, p.Name(), "pending")
+		pendingBuilds, err := ioutil.ReadDir(pendingPath)
+		for _, pending := range pendingBuilds {
+			pendingBuildPath := filepath.Join(pendingPath, pending.Name())
+			err = curfs.Remove(pendingBuildPath)
+			if err != nil {
+				return fmt.Errorf("Error pruning zombie build '%s' of project '%s'", pending.Name(), p.Name())
+			}
+			l.Printf("Pruned zombie build '%s' of project '%s'", pending.Name(), p.Name())
+		}
+	}
+
+	return nil
 }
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -109,3 +117,27 @@ func main() {
 		log.Fatal(err)
 	}
 }
+
+func cleanupPendingBuilds(curfs FileSystem) error {
```
agis: How about we call these zombie builds? Maybe `PruneZombieBuilds` is a better name?

main.go
```diff
@@ -109,3 +117,27 @@ func main() {
...
+	l := log.New(os.Stderr, "[cleanup] ", log.LstdFlags)
+	projects, err := ioutil.ReadDir(cfg.ProjectsPath)
+	if err != nil {
+		return err
+	}
+
+	for _, f := range projects {
```
agis: `p` (for "path" or "project") seems more accurate

main.go
```diff
@@ -109,3 +117,27 @@ func main() {
...
+		pendingPath := filepath.Join(cfg.BuildPath, f.Name(), "pending")
+		pendingBuilds, err := ioutil.ReadDir(pendingPath)
```
agis: How about just calling `os.RemoveAll()` and be done with it? `BootstrapProject()` will create the "pending" directory if it doesn't exist anyway.

agis: Nevermind my previous comment, we can't use `RemoveAll`, we must use `filesystem.Remove()`.

main.go
```diff
@@ -109,3 +117,27 @@ func main() {
...
+		if len(pendingBuilds) > 0 {
```
agis: This check is unnecessary.

main.go
```diff
@@ -109,3 +117,27 @@ func main() {
...
+			for _, p := range pendingBuilds {
+				l.Printf("Removing pending build of %s project: %s", f.Name(), p.Name())
```
agis: We could print this line after the directory was successfully deleted (and adapt the string accordingly).

agis: Something like:

> Pruned zombie build 'abc123' of project 'foo'

main.go
```diff
@@ -109,3 +117,27 @@ func main() {
...
+				pendingBuildPath := filepath.Join(pendingPath, p.Name())
+				err = curfs.Remove(pendingBuildPath)
+				if err != nil {
+					return err
```
agis: We should return a more meaningful error here:

> Error pruning zombie build abc123 of project '%s'

main.go
apostergiou: Yep, the directory deletion here should depend on the file system.

main.go
apostergiou: fixed

main.go
apostergiou: thanks for clarifying this, updated.

main.go
apostergiou: good catch, removed

main.go
apostergiou: true, changed.

main.go
apostergiou: Sure, I updated the name. :+1:

mistry_test.go
```diff
@@ -0,0 +1,39 @@
...
+	cmdOut, err := cliBuildJob("--project", project)
+	if err != nil {
+		t.Fatalf("Error output: %s, err: %v", cmdOut, err)
+	}
+	path := filepath.Join(cfg.BuildPath, project, "pending")
+	_, err = utils.RunCmd(curfs.Create(path + "/foo"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = utils.RunCmd(curfs.Create(path + "/bar"))
```
agis: Use `filepath.Join` to create the paths.

apostergiou: fixed

PR #13, opened 2018-04-03
Use project image in job ID seeding

agis:
This fixes the following bug:

- a job `{"project": "foo", "params": {"foo":"bar"}}` is built successfully
- a change to the Dockerfile (or any other file, for that matter) of project
  "foo" Dockerfile is applied
  - another job, identical to the first one arrives

Expected: the job should run after the new image is built (since the
Dockerfile changed).

Previous behavior: the cached results from the first execution were
returned, without actually re-building the image and re-running the job.

We solve this by using the project's contents (filenames and contents)
as a seed when generating the job ID.

### DIFF
--- a/end_to_end_test.go
+++ b/end_to_end_test.go
@@ -22,7 +22,7 @@ func TestSimpleBuild(t *testing.T) {
 }
 
 func TestUnknownProject(t *testing.T) {
-	expected := "Unknown project"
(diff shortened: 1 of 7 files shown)


agis:
Merged https://github.com/skroutz/mistry/commit/ef54d5c1bc0d2748a126dd6780e1232fc70ab62d.

PR #17, opened 2018-04-06
Add a Web view

apostergiou:
With this we add a web view for the service for monitoring the available jobs.

The feature introduces:
- an `index` page to view all available jobs and relevant information such as start time and state
- a `show` page for each where we can view the job result information alongside the job build logs. The show page provides a web tailf like feature for monitoring the build log as it unfolds.


tail_reader.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+type tailReader struct {
+	io.ReadCloser
+}
+
+// Read provides a tailf like generator.
+//
+// The reader sleeps on EOF and is shutdown when the file is closed.
+// Implements the io.Reader inerface.
```
agis: This line belongs on the struct documentation.

tail_reader.go
```diff
@@ -0,0 +1,39 @@
...
+		if n > 0 {
+			return n, nil
+		} else if err != io.EOF {
+			return n, err
+		}
+		time.Sleep(5 * time.Second)
+	}
+}
+
+func newTailReader(fileName string) (tailReader, error) {
```
agis: It's actually a `path`, not ` fileName`.

agis: It's idiomatic to return a pointer to the struct instead.

testdata/projects/simple-slow/docker-entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+sleep 30
+touch artifacts/out.txt
+
+exit 0
```
agis: This directory should be here only if it's actually used in mistry's tests.

types/build_result.go
```diff
@@ -23,6 +23,9 @@ type BuildResult struct {
 
 	// The method by which the build artifacts can be fetched.
 	TransportMethod TransportMethod
+
+	// The job's start time
+	StartTime string
```
agis: This should be a `time.Time`.

types/job_info.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+type JobInfo struct {
+	ID        string
+	Project   string
+	StartedAt time.Time
+	Output    string
+	Log       string
+	State     string
+}
```
agis: I see many duplicated fields between this and `Job`. Can we merge this into `Job`? It seems redundant.

server.go
```diff
@@ -22,6 +45,25 @@ type Server struct {
...
+
+	// Client connections registry.
+	//
+	// Cs sent to the newClients channel are being added to the registry.
+	// We use a reference to the C struct so that the conncetions can be
+	// uniquely identified for deletions.
+	clients map[*C]bool
+
+	// A connection queue used to track all open connections by their id.
+	cq map[string]bool
```
agis: All these fields and methods could go into a separate struct, ideally extracted into its own package and exporting its http handlers that one can use selectively.

tail_reader.go
```diff
@@ -0,0 +1,39 @@
...
+// Implements the io.Reader inerface.
+func (t tailReader) Read(b []byte) (int, error) {
+	for {
+		n, err := t.ReadCloser.Read(b)
+		if n > 0 {
+			return n, nil
+		} else if err != io.EOF {
+			return n, err
+		}
+		time.Sleep(5 * time.Second)
```
agis: Can you elaborate on why is this sleep needed?

job.go
```diff
@@ -192,3 +192,16 @@ func (j *Job) String() string {
 		"{project=%s params=%s group=%s id=%s}",
 		j.Project, j.Params, j.Group, j.ID[:7])
 }
+
+func jobPath(buildPath, project, id string) (string, string, error) {
```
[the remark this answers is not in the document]
apostergiou: Updated.

job.go
```diff
@@ -192,3 +192,16 @@ func (j *Job) String() string {
...
+	pPath := filepath.Join(buildPath, project, "pending", id)
+	rPath := filepath.Join(buildPath, project, "ready", id)
+	if _, err := os.Stat(pPath); err == nil {
```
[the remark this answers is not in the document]
apostergiou: Fixed.

local_assets.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+import "net/http"
+
+func staticFs() (http.FileSystem, error) {
+	return http.Dir("public"), nil
```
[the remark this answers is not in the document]
apostergiou: True, deleted.

public/index.html
```diff
@@ -0,0 +1,29 @@
...
+      </div>
+
+      <div class="large-12 cell" id="js-jobs">
+      </div>
+    </div>
+  </div>
+
+  <script src="https://unpkg.com/babel-standalone@6/babel.min.js"></script>
+  <script src="https://unpkg.com/react@16/umd/react.production.min.js"></script>
+  <script src="https://unpkg.com/react-dom@16/umd/react-dom.production.min.js"></script>
```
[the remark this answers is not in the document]
apostergiou: Good catch, moved inside the head tag.

public/js/index.js
```diff
@@ -0,0 +1,62 @@
...
+              </tr>
+            )
+           })}
+        </tbody>
+      </table>
+    );
+  }
+}
+
+ReactDOM.render(<Jobs every={5000} />, JobsRoot)
```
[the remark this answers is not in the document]
apostergiou: Ok, let's try with 3sec.

server.go
```diff
@@ -1,17 +1,40 @@
...
+	"strings"
 
 	"github.com/skroutz/mistry/types"
 )
 
+// C represents the connections to the job show page.
+//
+// Every struct instance will keep an open connection with the server
+// and listens for server side events.
+type C struct {
```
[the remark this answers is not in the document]
apostergiou: Updated.

tail_reader.go
```diff
@@ -0,0 +1,39 @@
+package main
+
+import (
+	"io"
+	"os"
+	"time"
+)
+
+type tailReader struct {
```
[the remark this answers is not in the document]
apostergiou: Done :+1: .

tail_reader.go
apostergiou: Fixed.

tail_reader.go
apostergiou: true, changed.

testdata/projects/simple-slow/docker-entrypoint.sh
apostergiou: Sure, removed.

tail_reader.go
apostergiou: Thanks for commenting on this. I added sleep here as a way to give up control to the scheduler while developing the feature. Nevertheless,  using it here feels like an anti-pattern and doesn't add anything so I removed it.

agis:
This PR should also update the relevant [README section](https://github.com/skroutz/mistry/commit/f8bac8d6e6c857ec5dd2b42cf42180f5340b3193#diff-04c6e90faac2675aa89e2176d2eec7d8R19).

server.go
apostergiou: Extracted into a `broker` package.

types/build_result.go
apostergiou: done

tail_reader.go
apostergiou: :+1:  done

README.md
```diff
@@ -16,7 +16,7 @@ Features include:
 - caching and reusing build results
 - efficient use of disk space due to copy-on-write semantics
 - a simple JSON API for interacting with the service
-- ([wip](https://github.com/skroutz/mistry/pull/17)) a web view for inspecting the progress and result of builds
+- a web view for inspecting the progress and result of builds
```
agis: This needs a rebase. I slightly updated this point.

job.go
```diff
@@ -207,3 +207,18 @@ func (j *Job) String() string {
 		"{project=%s params=%s group=%s id=%s}",
 		j.Project, j.Params, j.Group, j.ID[:7])
 }
+
+// State determines the job's current state by using it's path in the filesystem.
+func (j *Job) State(path, project, id string) (string, error) {
+	pPath := filepath.Join(path, project, "pending", id)
+	rPath := filepath.Join(path, project, "ready", id)
+	_, err := os.Stat(pPath)
+	if err == nil {
```
agis: What if error is non nil but also not a "not exist" error?

job.go
```diff
@@ -207,3 +207,18 @@ func (j *Job) String() string {
...
+		return "pending", nil
+	}
+	_, err = os.Stat(rPath)
+	if err == nil {
+		return "ready", nil
+	}
+	return "", err
```
agis: I don't like the fact that we may return an empty string and no error. What would such a case indicate?

job.go
```diff
@@ -207,3 +207,18 @@ func (j *Job) String() string {
 		"{project=%s params=%s group=%s id=%s}",
 		j.Project, j.Params, j.Group, j.ID[:7])
 }
+
+// State determines the job's current state by using it's path in the filesystem.
+func (j *Job) State(path, project, id string) (string, error) {
+	pPath := filepath.Join(path, project, "pending", id)
+	rPath := filepath.Join(path, project, "ready", id)
```
agis: Can't we use `j.PendingBuildPath` and `j.ReadyBuildPath`?

public/templates/show.html
```diff
@@ -0,0 +1,93 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>Job Logs</title>
+  <link rel="stylesheet" href="../../css/foundation.min.css">
+</head>
+
+<body>
+
+  <div class="top-bar">
+    <div class="top-bar-left">
+      <h1><a href="/">Mistry</a></h1>
```
agis: lowercase

public/templates/show.html
```diff
@@ -0,0 +1,93 @@
...
+      }
+
+      const source = new EventSource('/log/{{.Project}}/{{.ID}}');
+      const jobLog = document.getElementById('js-job-log')
+      source.onmessage = function(e) {
+        jobLog.innerHTML += e.data;
+      };
+
+      source.onerror = function(e) {
+        document.body.innerHTML += "Web tail failed.";
```
agis: Can we display more information about the error?

server_test.go
```diff
@@ -65,3 +72,87 @@ func TestLoad(t *testing.T) {
 		<-results
 	}
 }
+
+func TestIndex(t *testing.T) {
```
agis: Tests functions are named according to the functions they test. This should be named `TestHandleIndex` for example.

server_test.go
```diff
@@ -65,3 +72,87 @@ func TestLoad(t *testing.T) {
...
+	body, err := ioutil.ReadAll(result.Body)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !strings.Contains(string(body), expected) {
+		t.Errorf("Expeced body to contain %v, got %v", expected, string(body))
+	}
+}
+
+func TestShow(t *testing.T) {
```
agis: Ditto here.

server_test.go
```diff
@@ -65,3 +72,87 @@ func TestLoad(t *testing.T) {
...
+	// Build a job.
```
agis: Redundant.

types/build_result.go
```diff
@@ -23,6 +24,9 @@ type BuildResult struct {
 
 	// The method by which the build artifacts can be fetched.
 	TransportMethod TransportMethod
+
+	// The job's start time
+	StartTime time.Time
```
agis: This belongs in `Job`.

types/job_info.go
agis: Bump.

agis:
Please also add a small section in the README on where the web view can be found.

job.go
```diff
@@ -207,3 +207,18 @@ func (j *Job) String() string {
...
+	_, err := os.Stat(pPath)
```
agis: Instead for checking through the filesystem, which is racy if other methods of mistry manipulate the same paths, you can use the `JobQueue` struct which the server already contains.

agis:
We should add a new issue tracking what we need to revamp in the implementation.

types/job_info.go
apostergiou: Done.

README.md
apostergiou: True, fixed.

public/templates/show.html
apostergiou: done

server_test.go
apostergiou: Fixed

server_test.go
apostergiou: Fixed

server_test.go
apostergiou: Removed

types/build_result.go
apostergiou: Moved to `Job`.

job.go
apostergiou: It will return the error.

job.go
apostergiou: Good catch, but while I agree that this is racy, what would happen when we restart the server?The `JobQueue` would be empty then and the index page wouldn't display any jobs even if there are some. That's why I believe we should scan the filesystem.

job.go
agis: Correction: we do return a non-nil error here, but still, what does an empty string and a "ready path doesn't exist" error means?

job.go
apostergiou: I don't think we can as the job has been initialized with an empty struct 
```go
j := &Job{}
state, err := j.GetState(s.cfg.BuildPath, project, id)
```

Probably we can introduce a `GetJob` function that would find a job, WDYT?

public/templates/show.html
agis: https://developer.mozilla.org/el/docs/Web/JavaScript/Reference/Global_Objects/Error#Examples

job.go
agis: You're right, nvm.

job.go
agis: Let's revisit this at the next iteration of the web view. It's fine for now.

job.go
apostergiou: Thanks for noticing, maybe we could return a custom error then in which we have concatenated any errors with a meaningful message, e.g. "state not found for job %s error %v"

agis: A simple `"job %s not found error", j` I think is sufficient.

job.go
agis: Didn't notice that. Jobs should not be initialized this way, we should only use `NewJob()`.

job.go
apostergiou: Done

apostergiou:
> We should add a new issue tracking what we need to revamp in the implementation.

Tracked on #38.

PR #34, opened 2018-04-17
client: Pass --delete to rsync

apostergiou:
With this change, we delete extraneous files from the destination dir.

The produced command is: 
```shell
rsync --delete -rtlp apostergiou@localhost::mistry/simple/ready/c8d654f58623063cb55e0790890ad5aada1235be093ced09f7057e366b2de264/data/artifacts/* /tmp/foo

# client request

$ ./mistry-cli build --host localhost --port 8462 --project simple --group $RANDOM --target /tmp/foo --transport rsync -- --flavor=$RANDOM --x=$RANDOM
```

### DIFF
--- a/client/transport.go
+++ b/client/transport.go
@@ -27,5 +27,5 @@ func (ts Rsync) Copy(user, host, project, src, dst string) []string {
 	}
 	src = src[idx:]
 
-	return []string{"rsync", "-rtlp", fmt.Sprintf("%s@%s::%s/%s", user, host, module, src), dst}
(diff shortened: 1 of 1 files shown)


agis:
This is not urgent, so I think we should solve it just by making the client transport flags configurable.

apostergiou:
> This is not urgent, so I think we should solve it just by making the client transport flags configurable.

True, I opened an issue to track it https://github.com/skroutz/mistry/issues/35, so we can close this pull request.

PR #46, opened 2018-04-20
Fix logic to determine job directory bootstraping

dtheodor:
* cleanup logic: only job invocations with a group and a previous
build copy the previous build, the rest create new directories
* don't do unneccessary symlink lookup for non-group jobs
* add test case

- [x] update logic
- [x] factor out some functions

### DIFF
--- a/cmd/mistryd/job.go
+++ b/cmd/mistryd/job.go
@@ -20,6 +20,7 @@ import (
 	"github.com/docker/docker/api/types/mount"
 	docker "github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/stdcopy"
+	"github.com/skroutz/mistry/pkg/filesystem"
 	"github.com/skroutz/mistry/pkg/types"
 	"github.com/skroutz/mistry/pkg/utils"
 )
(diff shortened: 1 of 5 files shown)


dtheodor:
I added a new test project (`simple2`) which is identical to `simple`. I can't reuse `simple` in the new test case because we don't have granular test cleanup, and the output of the `simple` project is cleaned up at the end of the whole test suite. The new test case depends on starting with clean project data.

cmd/mistryd/worker.go
```diff
@@ -261,6 +221,55 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildResult *types.BuildResu
...
+	// if we cloned, empty the params dir
+	if cloneSrc != "" {
+		err = os.RemoveAll(filepath.Join(j.PendingBuildPath, DataDir, ParamsDir))
+		if err != nil {
+			err = workErr("could not remove params dir", err)
+			return
+		}
+	}
+
+	dirs := [4]string{
```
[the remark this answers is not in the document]
dtheodor: No need to pay the slice object allocation here, this array is only used to facilitate the for loop right below it and has a fixed known length

dtheodor:
Addressed comments. moved the logic to the `Job` struct

cmd/mistryd/job.go
```diff
@@ -207,3 +208,74 @@ func (j *Job) String() string {
 		"{project=%s params=%s group=%s id=%s}",
 		j.Project, j.Params, j.Group, j.ID[:7])
 }
+
+// CloneSrcPath Determine if we should use the build cache and return the path that
```
agis: s/Determine/determines/

cmd/mistryd/job.go
```diff
@@ -207,3 +208,74 @@ func (j *Job) String() string {
...
+// should be used.
+//
+// Using the build cache should happen when
+// 1. the job is invoked with a group
+// 2. the symlink pointing to the latest build is valid
+func (j *Job) CloneSrcPath(log *log.Logger) string {
```
agis: As a future idea, we should probably add a `logger` field to `Job{}` and use it throughout `Work()` and this function too.

cmd/mistryd/worker.go
```diff
@@ -89,48 +89,9 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildResult *types.BuildResu
...
-		if out != "" {
-			log.Println(out)
-		}
-		if err != nil {
-			err = workErr("could not create pending build path", err)
-			return buildResult, err
-		}
+	log.Printf("Creating new build directory...")
+	shouldCleanup, err := j.BootstrapBuildDir(s.cfg.FileSystem, log)
+	if shouldCleanup == true {
```
agis:
```
if shouldCleanup
```

cmd/mistryd/job.go
```diff
@@ -207,3 +208,74 @@ func (j *Job) String() string {
...
+				log.Printf("no latest build was found: %s", symlinkErr)
+			} else {
+				log.Printf("could not read latest build link, error: %s", symlinkErr)
+			}
+		}
+	}
+	return cloneSrc
+}
+
+// BootstrapBuildDir Create all required build directories. Returns true if directories have
```
agis: s/Create/creates/

Issue #25, opened 2018-04-12
Strategy for pruning old builds


agis:
There should be a strategy for somehow pruning old builds. This could be in the form of a "retention period" configuration.

This is necessary, disk usage will grow indefinitely.

agis:
We now have a script for doing this. Ideally, we want the server to do this automatically. One should be able to configure how much usage each project should consume and how much total usage all projects should consume.

PR #55, opened 2018-04-24
Add purger script for removing stale jobs

apostergiou:
A best-effort, temporary solution to #25.

The script will be replaced later when we implement a solution built-in to mistryd.

### DIFF
--- a/contrib/mistry-purge-builds
+++ b/contrib/mistry-purge-builds
@@ -1,0 +1,47 @@
+#!/usr/bin/env ruby
+require 'json'
+require 'time'
+require 'optparse'
+
+options = {}
+OptionParser.new do |opts|
+  opts.banner = "Purge old mistry builds from the file system.\nUsage: #{$0} [options]"
+  opts.on('--older-than DAYS' 'remove builds older than DAYS days') { |v| options[:stale_point] = Time.now - Integer(v)*24*60*60 }
+  opts.on('--path PATH', 'Build path') { |v| options[:path] = v }
+  opts.on('--dry-run', 'Dry run') { |v| options[:dry_run] = v }
+end.parse!
+
+abort("#{options[:path]} is not a directory") unless File.directory?(options[:path])
+
+jobs = []
+stale_jobs = []
+groups_and_latest = []
+projects = Dir["#{options[:path]}/projects/*"].map { |p| p.gsub("#{options[:path]}/projects/","") }
+
+projects.each do |p|
+  data_path = File.join(options[:path], "data", p)
+  groups_path = File.join(data_path, "groups")
+  latest_path = File.join(data_path, "latest")
+  # jobs that have a symlink either as a group or as latest won't be deleted
+  group_jobs = Dir["#{groups_path}/*"]
+  groups_and_latest << File.readlink(Dir["latest_path"]) if Dir["latest_path"].first
+  group_jobs.each do |j|
+    groups_and_latest << File.readlink(j)
+  end
+
+  ready_path = File.join(data_path, "ready")
+  ready_jobs = Dir["#{ready_path}/*"]
+  ready_jobs.each do |rj|
+    if t = JSON.parse(File.read("#{rj}/build_info.json"))["StartedAt"]
+      start_time = Time.parse(t)
+    end
+    stale_jobs << rj if start_time.nil? || start_time < options[:stale_point]
+  end
+end
+
+if options[:dry_run]
+  puts "would delete jobs: #{stale_jobs}"
+elsif !stale_jobs.empty?
+  `btrfs subvolume delete #{stale_jobs.join(' ')}`
+  File.unlink(*(groups_and_latest.select{ |j| stale_jobs.include?(File.readlink(j)) }))
+end


contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
...
+stale_jobs = []
+
+projects.each do |p|
+  data_path = File.join(options[:path], "data", p)
+  groups_path = File.join(data_path, "groups")
+  latest_path = File.join(data_path, "latest")
+  # jobs that have a symlink either as a group or as latest won't be deleted
+  groups_and_latest = []
+  group_jobs = Dir["#{groups_path}/*"]
+  groups_and_latest << `readlink #{Dir["latest_path"]}`
```
ctrochalakis: You can use `File.readlink()` and avoid the shell escape.

contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
...
+  group_jobs.each do |j|
+    groups_and_latest << `readlink #{j}`.strip
+  end
+
+  ready_path = File.join(data_path, "/ready")
```
ctrochalakis: nit, you should drop the slash. `File.join` is used to avoid hardcoding the file seperator (`/`) for portability (e.g. on windows systems)

contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
...
+  ready_jobs = Dir["#{ready_path}/*"]
+  (ready_jobs - groups_and_latest).each do |rj|
+    if t = JSON.parse(File.read("#{rj}/build_info.json"))["StartedAt"]
```
ctrochalakis: Convert to `File.join()`

ctrochalakis:
We have jobs that don't have a buildinfo file,right? We should have a plan for them.

contrib/build_purger.rb
apostergiou: done

contrib/build_purger.rb
apostergiou: good catch, fixed

contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env ruby
+require 'json'
+require 'time'
+require 'optparse'
+require 'yaml'
+
+options = {}
+OptionParser.new do |opts|
+  opts.banner = "Usage: build_purger.rb [options]"
```
agis: We could add a small description here so that the user knows what this is about. Something like:

> Purge old mistry builds from the file system.

apostergiou:
> We have jobs that don't have a buildinfo file,right? We should have a plan for them.

Jobs that don't have the buildinfo file will be considered stale and eventually will be deleted, except those that are the symbolic link target of `groups` and `latest`.

ctrochalakis:
> Jobs that don't have the buildinfo file will be considered stale and eventually will be deleted, except those that are the symbolic link target of groups and latest.

Not sure about that, I think it's better to delete them (we should make sure that it doesn't affect the server readlink logic). The problem is that this way we will never delete obsolete groups and reclaim their space.

agis:
Please describe in the commit message (and PR) why this change is needed (ie. that it's a best-effort, temporary solution to #25) and that it will be replaced later when we implement a solution built-in to mistryd.

contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
...
+  opts.on('--days DAYS', 'Stale point') { |v| options[:stale_point] = Time.now - v.to_i*24*60*60 }
```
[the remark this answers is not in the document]
apostergiou: done

contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+    stale_jobs << rj if start_time.nil? || start_time < options[:stale_point]
+  end
+end
+
+if options[:dry_run]
+  puts "would delete jobs: #{stale_jobs}"
+else
+  stale_jobs.each do |sj|
+    `btrfs subvolume delete #{sj}`
```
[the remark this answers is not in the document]
apostergiou: Great catch, thank you for pointing this out! The code has been fixed to play with multiple arguments.

contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
apostergiou: dropped the extension and renamed

contrib/build_purger.rb
```diff
@@ -0,0 +1,49 @@
...
+require 'json'
+require 'time'
+require 'optparse'
+require 'yaml'
```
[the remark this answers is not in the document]
apostergiou: nope, removed. I added it as part of the experimentation with a static yaml file for the commands but it was overkill.

contrib/build_purger.rb
apostergiou: done

apostergiou:
Comments addressed and updated PR description.

contrib/mistry-purge-builds
```diff
@@ -0,0 +1,47 @@
...
+      start_time = Time.parse(t)
+    end
+    stale_jobs << rj if start_time.nil? || start_time < options[:stale_point]
+  end
+end
+
+if options[:dry_run]
+  puts "would delete jobs: #{stale_jobs}"
+else
+  `btrfs subvolume delete #{stale_jobs.join(' ')}`
```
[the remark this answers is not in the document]
apostergiou: Unless I'm very much mistaken, the command will print an error. Added a simple condition to mitigate this.

contrib/mistry-purge-builds
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env ruby
+require 'json'
+require 'time'
+require 'optparse'
+
+options = {}
+OptionParser.new do |opts|
+  opts.banner = "Purge old mistry builds from the file system.\nUsage: mistry-purge-builds [options]"
```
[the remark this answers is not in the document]
apostergiou: True, thanks :+1:

PR #58, opened 2018-04-25
Change filesystem Create and Clone, and Btrfs.Remove behavior

dtheodor:
* Btrfs.Remove no longer returns an error if the path does not exist, similar to the Plainfs behavior. The interface for Remove was updated to document that
* Create and Clone now execute the operation instead of returning a
command string array
* rename filesystem.List to filesystem.Registry

### DIFF
--- a/pkg/filesystem/btrfs/btrfs.go
+++ b/pkg/filesystem/btrfs/btrfs.go
@@ -2,6 +2,7 @@ package btrfs
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/skroutz/mistry/pkg/filesystem"
 	"github.com/skroutz/mistry/pkg/utils"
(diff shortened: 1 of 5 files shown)


pkg/filesystem/btrfs/btrfs.go
```diff
@@ -13,19 +14,27 @@ import (
...
 func (fs Btrfs) Remove(path string) error {
-	out, err := utils.RunCmd([]string{"btrfs", "subvolume", "delete", path})
+	_, err := os.Stat(path)
+	if err == nil {
+		return runCmdWithBetterError([]string{"btrfs", "subvolume", "delete", path})
+	}
+	return nil
+}
+
+func runCmdWithBetterError(args []string) error {
```
agis: A simple `runCmd` would perhaps be better, since the body of the function is very simple.

pkg/filesystem/plainfs/plainfs.go
```diff
@@ -4,24 +4,29 @@ import (
...
+func (fs PlainFS) Create(path string) error {
+	return os.Mkdir(path, 0755)
 }
 
-func (fs PlainFS) Clone(src, dst string) []string {
-	return []string{"cp", "-r", src, dst}
+// Clone recursively copies the contents of src to dst
+func (fs PlainFS) Clone(src, dst string) error {
+	_, err := utils.RunCmd([]string{"cp", "-r", src, dst})
+	return err
```
agis: Is there a reason we don't also use `runCmdWithBetterError` here?

PR #61, opened 2018-04-27
Trigger a build asynchronously

### LINKED ISSUE #15: Support non-blocking job schedule
Sometimes, it's necessary to just trigger a job and not wait for it to finish.

A common scenario is when someone wants to eagerly prepare a build that knows is going to be needed in the future (eg. triggering a build from a github webhook to precompile the assets of a Rails application).

The server should support this option first, and then it has to be implemented in the client.

dtheodor:
Add a `mistry build --no-wait` flag that triggers a build and exits without waiting

* add a new client flag
* update the server endpoint to accept `POST /jobs?async`, returns an empty 201 response on success
* update README and CLI docs

Some small random improvements
* use {{.HelpName}}  and {{.Name}} in help templates to avoid hardcoding command names
* add some comments to exported functions to stop go fmt complains
* ~embed the JobRequest into the Job to avoid field declaration duplication~

Closes #15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Breaking changes are prefixed with a "[BREAKING]" label.
 
 ### Added
 
+- Asynchronous job scheduling [[#61](https://github.com/skroutz/mistry/pull/61)]
 - Web view [[#17](https://github.com/skroutz/mistry/pull/17)]
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -91,21 +91,45 @@ using the JSON API directly. We recommended using `mistry` whenever possible
 (although it may occassionally lag behind the server in terms of
 supported features).
 
-**Scheduling a new job and fetching the artifacts:**
-```shell
-$ mistry --project foo
+#### Client
+
+##### Schedule a new job, wait for completion, and fetch the artifacts
+
+```sh
+$ mistry build --project foo --target .
 ```
-This will place the artifacts and the current working directory.
+This will place the artifacts and the target directory.
+
+##### Schedule a new job without waiting
+
+```sh
+$ mistry build --project foo --async
+```
 
 See `mistry build -h` for more options.
 
-**Scheduling a new job without fetching the artifacts**:
-``` shell
-$ curl -H 'Content-Type: application/json' -d '{"project": "foo"}' localhost:8462/jobs
+#### HTTP Endpoints
+
+##### Schedule a new job
+
+```sh
+$ curl -X POST /jobs \
+    -H 'Accept: application/json' \
+    -H 'Content-Type: application/json' \
+    -d '{"project": "foo"}' 
 ```
 
-
-
+```js
+{
+    "Params": {"foo": "xzv"}, 
+    "Path": "<artifact path>",
+    "Cached": true,
+    "Coalesced": false,
+    "ExitCode": 0,
+    "Err": null,
+    "TransportMethod": "rsync"
+}
+```
 
 
 ### Web view
--- a/cmd/mistry/main.go
+++ b/cmd/mistry/main.go
@@ -51,6 +51,7 @@ func main() {
 		transport     string
 		verbose       bool
 		jsonResult    bool
+		noWait        bool
 	)
 
 	currentUser, err := user.Current()
@@ -67,29 +68,27 @@ JOB PARAMETERS:
    -- dynamic options for the command
 
 EXAMPLES:
-   1. The following sequence schedules a command to be executed in http://example.org:9090,
-      builds the yarn project, groups it under group_name, syncs the result to the
-      directory /tmp/yarn and uses the dynamic arguments yarn.lock and foo.
-      Note the @ before the yarn.lock, this indicates a referral to an actual file on the
-      filesystem.
+	1. Schedule a job with a group and some parameters and put artifacts under
+		/tmp/yarn using rsync. Prefixing a file name with @ will cause the contents
+		of yarn.lock to be sent as parameters.
 
-      $ mistry-cli build --host example.org --port 9090 --project yarn \
-      --group group_name --target /tmp/yarn -- --lockfile=@yarn.lock --foo=bar
+		$ {{.HelpName}} --host example.org --port 9090 --project yarn \
+			--group group_name --transport rsync --target /tmp/yarn \
+			-- --lockfile=@yarn.lock --foo=bar
 
-   2. The following sequence uses the GROUP environment variable to set the project group.
+	2. Schedules a build and exits early without waiting for the result by setting
+		the no-wait flag
 
-      $ GROUP=group_name mistry-cli build --host example.org --port 9090 --project yarn \
-      --target /tmp/yarn -- --lockfile=@yarn.lock --foo=bar
-	`, cli.CommandHelpTemplate)
+		$ {{.HelpName}} --host example.org --port 9090 --project yarn --no-wait
+`, cli.CommandHelpTemplate)
 
 	app := cli.NewApp()
-	app.Name = "mistry-cli"
-	app.Usage = "mistry client"
+	app.Usage = "schedule build jobs at the mistry service"
 	app.HideVersion = true
 	app.Commands = []cli.Command{
 		{
 			Name:  "build",
-			Usage: "Schedule a mistry job and retrieve the results",
+			Usage: "Schedule jobs and retrieve artifacts.",
 			Flags: []cli.Flag{
 				cli.StringFlag{
 					Name:        "host",
@@ -104,12 +103,6 @@ EXAMPLES:
 					Value:       "8462",
 				},
 				cli.StringFlag{
-					Name:        "transport-user",
-					Usage:       "user to fetch the artifacts with",
-					Destination: &transportUser,
-					Value:       currentUser.Username,
-				},
-				cli.StringFlag{
 					Name:        "project",
 					Usage:       "job's project",
 					Destination: &project,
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -91,21 +91,45 @@ using the JSON API directly. We recommended using `mistry` whenever possible
...
-**Scheduling a new job without fetching the artifacts**:
-``` shell
-$ curl -H 'Content-Type: application/json' -d '{"project": "foo"}' localhost:8462/jobs
+```sh
+$ mistry build --project foo --async
 ```
 
+See `mistry build -h` for more options.
 
+#### HTTP Endpoints
```
agis: I believe we should place this under a wiki page and gradually enrich it.

cmd/mistry/main.go
```diff
@@ -67,29 +68,34 @@ JOB PARAMETERS:
...
-   1. The following sequence schedules a command to be executed in http://example.org:9090,
-      builds the yarn project, groups it under group_name, syncs the result to the
-      directory /tmp/yarn and uses the dynamic arguments yarn.lock and foo.
-      Note the @ before the yarn.lock, this indicates a referral to an actual file on the
-      filesystem.
+    1. Talk to the mistry service at http://example.org:9090, to build the yarn
+       project, group it under group_name, sync the result to the
+       directory /tmp/yarn and use the dynamic arguments yarn.lock and foo.
+       Note the @ before the yarn.lock, this indicates a referral to an actual file on the
+       filesystem.
```
agis: I think we can simplify this even more:

> Schedule a job with a group and some parameters and put artifacts under `/tmp/yarn` using rsync. Prefixing a file name with @ will cause the contents of yarn.lock to be sent as parameters.

cmd/mistry/main.go
```diff
@@ -67,29 +68,34 @@ JOB PARAMETERS:
...
 
-      $ mistry-cli build --host example.org --port 9090 --project yarn \
-      --group group_name --target /tmp/yarn -- --lockfile=@yarn.lock --foo=bar
+       $ {{.HelpName}} --host example.org --port 9090 --project yarn \
+	     --group group_name --transport rsync --target /tmp/yarn \
+	     -- --lockfile=@yarn.lock --foo=bar
 
-   2. The following sequence uses the GROUP environment variable to set the project group.
+    2. Uses the GROUP environment variable to set the project group.
```
agis: Since we're at it, I think this example isn't needed at all. A note that says that every flag can be provided as environment variable too should be sufficient.

cmd/mistry/main.go
```diff
@@ -67,29 +68,34 @@ JOB PARAMETERS:
...
-      $ GROUP=group_name mistry-cli build --host example.org --port 9090 --project yarn \
-      --target /tmp/yarn -- --lockfile=@yarn.lock --foo=bar
-	`, cli.CommandHelpTemplate)
+	   $ GROUP=group_name {{.HelpName}} --host example.org --port 9090 --project yarn \
+	     --transport rsync --target /tmp/yarn -- --lockfile=@yarn.lock --foo=bar
+
+    3. Schedules a build and exits early without waiting for the result by not
+	   specifying a transport
+
+       $ {{.HelpName}} -host example.org --port 9090 --project yarn
```
agis: Need update.

cmd/mistry/main.go
```diff
@@ -67,29 +68,34 @@ JOB PARAMETERS:
...
+`, cli.CommandHelpTemplate)
 
 	app := cli.NewApp()
-	app.Name = "mistry-cli"
```
agis: Is the `.Name` set automatically by urfave/cli?

cmd/mistry/main.go
```diff
@@ -67,29 +68,34 @@ JOB PARAMETERS:
...
-	app.Usage = "mistry client"
+	app.Usage = "schedule build jobs at the mistry service"
 	app.HideVersion = true
 	app.Commands = []cli.Command{
 		{
 			Name:  "build",
-			Usage: "Schedule a mistry job and retrieve the results",
+			Usage: "Schedule a mistry job, wait for it to finish, and retrieve the results",
```
agis: Since this is not always the case, we should just say something like:

> Schedule jobs and retrieve artifacts.

cmd/mistry/main.go
```diff
@@ -139,6 +128,31 @@ EXAMPLES:
...
+
+				// transport flags
+				cli.BoolFlag{
+					Name:        "async",
+					Usage:       "If set, don't wait for the build to finish and exit once the build is scheduled",
+					Destination: &async,
+				},
+				cli.StringFlag{
+					Name:        "transport",
+					Usage:       "Specify the transport method to transfer the build results",
```
agis: Suggestion:

> The method to use for fetching artifacts.

cmd/mistry/main.go
```diff
@@ -139,6 +128,31 @@ EXAMPLES:
 					Usage:       "Output the build result in JSON format to STDOUT (implies verbose: false)",
 					Destination: &jsonResult,
 				},
+
+				// transport flags
+				cli.BoolFlag{
+					Name:        "async",
+					Usage:       "If set, don't wait for the build to finish and exit once the build is scheduled",
```
agis: Suggestion:

> If set, schedule the job but don't fetch the artifacts.

cmd/mistry/main.go
```diff
@@ -139,6 +128,31 @@ EXAMPLES:
 					Usage:       "Output the build result in JSON format to STDOUT (implies verbose: false)",
 					Destination: &jsonResult,
 				},
+
+				// transport flags
+				cli.BoolFlag{
+					Name:        "async",
```
agis: I think the term "async" carries more weight than we want to give to this option. How about something like:

- no-wait
- no-fetch
- no-artifacts
- skip-artifacts
- no-block

cmd/mistry/main.go
```diff
@@ -139,6 +128,31 @@ EXAMPLES:
...
+				},
+				cli.StringFlag{
+					Name:        "transport-user",
+					Usage:       "user to fetch the artifacts with",
+					Destination: &transportUser,
+					Value:       currentUser.Username,
+				},
+				cli.StringFlag{
+					Name:        "target, t",
+					Usage:       "the local directory where the result will be saved",
```
agis: Let's be consistent with our glossary and use the term "artifacts" instead of "results".

cmd/mistry/main.go
```diff
@@ -139,6 +128,31 @@ EXAMPLES:
...
+				},
+				cli.StringFlag{
+					Name:        "transport",
+					Usage:       "Specify the transport method to transfer the build results",
+					Destination: &transport,
+					Value:       types.Scp,
+				},
+				cli.StringFlag{
+					Name:        "transport-user",
+					Usage:       "user to fetch the artifacts with",
```
agis: [nit] for consistency, s/user/User

cmd/mistry/main.go
agis: [nit] also for consistency, s/the/The

cmd/mistry/main.go
```diff
@@ -151,15 +165,26 @@ EXAMPLES:
...
 				if jsonResult {
 					verbose = false
 				}
 
+				var (
+					ts       Transport
+					tsExists bool
+				)
+
+				if transport != "" {
```
agis: This check is redundant.

cmd/mistry/main.go
```diff
@@ -187,40 +212,24 @@ EXAMPLES:
 					}
 				}
 
-				jr := types.JobRequest{Project: project, Group: group, Params: params}
-				jrJSON, err := json.Marshal(jr)
-				if err != nil {
-					return err
+				url := fmt.Sprintf("http://%s:%s/%s", host, port, JobsPath)
+				if async {
+					url += "?async"
```
agis: Shouldn't we better use `async=true`? I think semantically it's more proper.

cmd/mistry/main.go
```diff
@@ -187,40 +212,24 @@ EXAMPLES:
 					}
 				}
 
-				jr := types.JobRequest{Project: project, Group: group, Params: params}
-				jrJSON, err := json.Marshal(jr)
-				if err != nil {
-					return err
+				url := fmt.Sprintf("http://%s:%s/%s", host, port, JobsPath)
```
agis: Note: this hardcoded protocol might prove a problem in the future.

cmd/mistry/main.go
```diff
@@ -187,40 +212,24 @@ EXAMPLES:
...
+				if async {
+					url += "?async"
 				}
 
-				request, err := http.NewRequest("POST", fmt.Sprintf("http://%s:%s/%s", host, port, JobsPath), bytes.NewBuffer(jrJSON))
+				body, err := sendRequest(url, verbose, project, group, params)
```
agis: Maybe better construct the `JobRequest` before calling this function and just pass it as a parameter? The function would then would only do what it's supposed to.

cmd/mistryd/end_to_end_test.go
```diff
@@ -34,6 +34,15 @@ func TestNonGroupSubsequentInvocation(t *testing.T) {
 	}
 }
 
+func TestAsyncSimpleBuild(t *testing.T) {
+	cmdout, cmderr, err := cliBuildJob("--project", "bootstrap-twice", "--async")
+	if err != nil {
+		t.Fatalf("mistry-cli stdout: %s, stderr: %s, err: %#v", cmdout, cmderr, err)
+	}
+	assertEq(cmdout, "", t)
+	assertEq(cmderr, "", t)
```
agis: I think we should test here that the build has indeed run successfully.

cmd/mistryd/main.go
```diff
@@ -57,6 +57,7 @@ const (
 	// Docker images/containers created by mistry.
 	ImgCntPrefix = "mistry-"
 
+	// DateFmt the date format
```
agis: s/DateFmt//

cmd/mistryd/job.go
```diff
@@ -32,9 +32,7 @@ type Job struct {
 	ID string
 
 	// user-provided
-	Project string
-	Params  types.Params
-	Group   string
```
agis: `Job` is a core struct and should not get everything that a `JobRequest` has. In the future we might add more fields and methods to `JobRequest` but we don't want to add them here.

On the other hand, I see why you chose to do so and perhaps we should extract those common fields into yet another type called `JobInfo` maybe or something like that.

Anyway, I think we should leave this for another patch.

cmd/mistryd/server.go
```diff
@@ -134,6 +144,12 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) handleNewJobAsync(j *Job, w http.ResponseWriter) {
+	s.Log.Printf("Scheduling %s...", j)
+	defer s.Work(context.Background(), j)
```
agis: We don't want to leave the connection open. You could just `go s.Work` this.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Breaking changes are prefixed with a "[BREAKING]" label.
 
 ### Added
 
+- Asynchronous build scheduling [[#61](https://github.com/skroutz/mistry/pull/61)]
```
agis: s/build/job/

cmd/mistry/main.go
dtheodor: yes, `os.Args[0]`

https://github.com/urfave/cli/blob/3fb51f15e8ce6ea8428a8dacb30d120fad8fa186/app.go#L83

cmd/mistryd/server.go
dtheodor: yes, I meant to go this

cmd/mistry/main.go
dtheodor: not necessarily. URL params can support semantics similar to CLI boolean flags, i.e. the existence of the flag `--async` is what determines if it is true, absense of it determines it to be false. similarly in URL params, existence of the param `?async` or lack of it determines its value. unvalued URL params are supported by the spec

giving it an explicit value `?async=true` only makes sense if this is a three-valued boolean, i.e. if the absence of it also has meaning and we want to support the three values `true` `false` and `null`. we don't have this need here

agis: Makes sense!

cmd/mistry/main.go
dtheodor: lets use a single parameter `--server-url http://example.org:834`

agis: That'd make sense, but let's do it some other time. Want to open an issue to track this?

agis:
LGTM!

PR #62, opened 2018-04-27
rerun failed builds instead of using their cached result

### LINKED ISSUE #45: Consider not caching failed builds
I think that it makes more sense not to cache build failures, most of the time those would be transient errors that should be retried next time someone asks for that build.

In that case we might want to keep the build directory (in a `failed` subdirectory?) for debugging purposes.

dtheodor:
Re-run "ready" jobs whose exit status is not 0. Re-running such a job means removing the existing ready folder and then proceeding as usual, i.e. creating a new pending folder etc.

Closes #45

### DIFF
--- a/cmd/mistryd/end_to_end_test.go
+++ b/cmd/mistryd/end_to_end_test.go
@@ -174,28 +174,33 @@ func TestResultCache(t *testing.T) {
 	assert(br2.ExitCode, 0, t)
 }
 
-func TestResultCacheExitCode(t *testing.T) {
-	cmdout1, _, err := cliBuildJob("--json-result", "--project", "result-cache-exitcode")
+func TestRerunFailedBuild(t *testing.T) {
+	// schedule a build that fails (non zero exit code)
+	cmdout, _, err := cliBuildJob("--json-result", "--project", "exit-code")
 	if err == nil {
-		t.Fatal("Expected error")
+		t.Fatal("expected error")
 	}
-	bi, err := parseClientJSON(cmdout1)
-	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/mistryd/end_to_end_test.go
```diff
@@ -176,28 +176,57 @@ func TestResultCache(t *testing.T) {
 
 func TestResultCacheExitCode(t *testing.T) {
```
agis: Does this test case makes sense at all after this change?

cmd/mistryd/end_to_end_test.go
```diff
@@ -176,28 +176,57 @@ func TestResultCache(t *testing.T) {
...
+	if err == nil {
+		t.Fatal("expected error")
+	}
+
+	br, err := parseClientJSON(cmdout)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assertNotEq(br.ExitCode, 0, t)
```
agis: Why the exit code is 0 here? Shouldn't it be 33?

dtheodor: we're testing whether the result was cached. a result is now only cached with an exit code of 0. so the test case now needs an exit code of 0

dtheodor: i see we have another test for that actually, so I'm going to remove this one

agis:
LGTM 👏

agis:
Oh, please add an entry in the changelog (under `Changed`). As long as the user is concerned, we a build that errors is not going to be cached.

PR #63, opened 2018-04-30
Add --clear-target option

### LINKED ISSUE #60: Add option to completely replace the target directory
In some scenarios, you want the artifacts fetched from mistry to completely override the previous contents of the `--target-path`. To support such use cases, we could add a `--replace` (or `replace-target`) boolea option that, if enabled will completely replace the given path.

In the case of rsync, it would just pass the `--delete` option to rsync.

dtheodor:
- [x] add --clear-target option, clears the --target directory
  * rsync: adds the `--delete` flag
  * scp: calls `os.Remove`s before running the `scp` command. `scp` has no flag that can do this
- [x] change Copy interface to do the copying and return `(string, error)` instead of returning the command string `[]string`. this is necessary to deal with clear-target errors, plus a design improvement as the Copy implementation should be an implementation detail (which could use an API instead of running commands on the OS), not the interface contract
- [x] improve test case setup code to allow to easily override the CLI default options (target directory in this case)

Closes #60

### DIFF
--- a/cmd/mistry/transport.go
+++ b/cmd/mistry/transport.go
@@ -7,19 +7,27 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/skroutz/mistry/pkg/utils"
 )
 
 type Transport interface {
-	Copy(user, host, project, src, dst string, clearDst bool) []string
+	// Copy copies from src to dst, clearing the destination if clearDst is true
+	Copy(user, host, project, src, dst string, clearDst bool) (string, error)
 }
 
 type Scp struct{}
 
-func (ts Scp) Copy(user, host, project, src, dst string, clearDst bool) []string {
+// Copy runs 'scp user@host:src dst'. If clearDst is set, all contents of dst will be
+// removed before the scp
+func (ts Scp) Copy(user, host, project, src, dst string, clearDst bool) (string, error) {
 	if clearDst {
-		removeDirContents(dst)
+		err := removeDirContents(dst)
+		if err != nil {
+			return "", err
+		}
 	}
-	return []string{"scp", "-r", fmt.Sprintf("%s@%s:%s", user, host, src), dst}
+	return utils.RunCmd([]string{"scp", "-r", fmt.Sprintf("%s@%s:%s", user, host, src), dst})
 }
 
 func removeDirContents(dir string) error {
@@ -39,7 +47,9 @@ func removeDirContents(dir string) error {
 
 type Rsync struct{}
 
-func (ts Rsync) Copy(user, host, project, src, dst string, clearDst bool) []string {
+// Copy runs 'rsync -rtlp user@host::mistry/src dst'. If clearDst is true, the --delete flag
+// will be set
+func (ts Rsync) Copy(user, host, project, src, dst string, clearDst bool) (string, error) {
 	module := "mistry"
 
 	idx := strings.Index(src, project)
@@ -53,5 +63,5 @@ func (ts Rsync) Copy(user, host, project, src, dst string, clearDst bool) []stri
 	}
 	cmd = append(cmd, fmt.Sprintf("%s@%s::%s/%s", user, host, module, src), dst)
 
-	return cmd
+	return utils.RunCmd(cmd)
 }
--- a/cmd/mistry/transport.go
+++ b/cmd/mistry/transport.go
@@ -2,23 +2,44 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
+	"os"
+	"path/filepath"
 	"strings"
 )
 
 type Transport interface {
-	Copy(user, host, project, src, dst string) []string
+	Copy(user, host, project, src, dst string, clearDst bool) []string
 }
 
 type Scp struct{}
 
-func (ts Scp) Copy(user, host, project, src, dst string) []string {
+func (ts Scp) Copy(user, host, project, src, dst string, clearDst bool) []string {
+	if clearDst {
+		removeDirContents(dst)
+	}
 	return []string{"scp", "-r", fmt.Sprintf("%s@%s:%s", user, host, src), dst}
+}
+
+func removeDirContents(dir string) error {
+	items, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return err
+	}
+
+	for _, item := range items {
+		err = os.RemoveAll(filepath.Join(dir, item.Name()))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 type Rsync struct{}
 
-func (ts Rsync) Copy(user, host, project, src, dst string) []string {
+func (ts Rsync) Copy(user, host, project, src, dst string, clearDst bool) []string {
 	module := "mistry"
 
 	idx := strings.Index(src, project)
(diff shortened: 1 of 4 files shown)


cmd/mistryd/end_to_end_test.go
```diff
@@ -89,6 +89,48 @@ func waitUntilExists(path string) error {
...
+		f, err := os.Create(filepath)
+		failIfError(err, t)
+		f.Close()
+	}
+
+	// run the job with remove-target
+	cmdout, cmderr, err := cliBuildJobTarget(target, "--project", "simple", "--verbose", "--clear-target")
+	t.Logf("cli output: out: %s err: %s", cmdout, cmderr)
+	if err != nil {
+		t.Fatalf("mistry-cli stdout: %s, stderr: %s, err: %#v", cmdout, cmderr, err)
```
apostergiou: If I am not mistaken we are using `failIfError` to exit immediately on special cases. You could use the function here, couldn't you?

In addition, if we are going to utilize this helper function we could refactor the tests (in another PR of course). @agis WDYT?

cmd/mistryd/end_to_end_test.go
```diff
@@ -89,6 +89,48 @@ func waitUntilExists(path string) error {
...
+	}
+	// verify the files and directory have been deleted
+	for _, path := range append(fileNames, dirName) {
+		_, err = os.Stat(path)
+		if err == nil {
+			t.Fatalf("unexpected file found at %s", path)
```
apostergiou: ditto

cmd/mistryd/end_to_end_test.go
```diff
@@ -89,6 +89,48 @@ func waitUntilExists(path string) error {
...
+		} else if !os.IsNotExist(err) {
+			t.Fatalf("error when trying to check target file %s: %s", path, err)
```
apostergiou: ditto

cmd/mistryd/end_to_end_test.go
```diff
@@ -89,6 +89,48 @@ func waitUntilExists(path string) error {
...
+
+	for _, filepath := range fileNames {
+		f, err := os.Create(filepath)
+		failIfError(err, t)
+		f.Close()
+	}
+
+	// run the job with remove-target
+	cmdout, cmderr, err := cliBuildJobTarget(target, "--project", "simple", "--verbose", "--clear-target")
+	t.Logf("cli output: out: %s err: %s", cmdout, cmderr)
```
apostergiou: :scissors:

cmd/mistryd/mistryd_test.go
```diff
@@ -246,11 +246,18 @@ func parseClientJSON(s string) (*types.BuildInfo, error) {
...
 // working directory where the tests are ran from.
 func cliBuildJob(args ...string) (string, string, error) {
+	return cliBuildJobTarget(target, args...)
+}
+
+func cliBuildJobTarget(trgt string, args ...string) (string, string, error) {
 	clientPath := os.Getenv("MISTRY_CLIENT_PATH")
 	if clientPath == "" {
-		clientPath = "./mistry"
+		clientPath = "/home/dtheodor/go/src/github.com/skroutz/mistry/mistry"
```
apostergiou: This code should stay independent of the dev path.

cmd/mistryd/end_to_end_test.go
```diff
@@ -107,9 +107,11 @@ func TestBuildRemoveTarget(t *testing.T) {
 		failIfError(err, t)
 		f.Close()
 	}
+	cliArgs := cliDefaultArgs
```
apostergiou: Nice refactoring :+1:

cmd/mistryd/mistryd_test.go
dtheodor: oops

cmd/mistryd/end_to_end_test.go
dtheodor: I would have to modify `failIfError` if we want to add a custom error message with interpolated variables

cmd/mistryd/end_to_end_test.go
dtheodor: i can't really use it here, `failIfError` deals with the common

```
if err != nil {
    t.Fatalf()
}
```

The error conditions here are different, they are special cases (`if err == nil` and `if os.IsNotExist(err)`)

cmd/mistry/transport.go
```diff
@@ -2,30 +2,63 @@ package main
...
+	"os"
+	"path/filepath"
 	"strings"
+
+	"github.com/skroutz/mistry/pkg/utils"
 )
 
 type Transport interface {
-	Copy(user, host, project, src, dst string) []string
+	Copy(user, host, project, src, dst string, clearDst bool) (string, error)
```
agis: Needs some documentation.

cmd/mistry/transport.go
```diff
@@ -2,30 +2,63 @@ package main
...
+		err := removeDirContents(dst)
+		if err != nil {
+			return "", err
+		}
+	}
+	out, err := utils.RunCmd([]string{"scp", "-r", fmt.Sprintf("%s@%s:%s", user, host, src), dst})
+	return out, err
+}
+
+func removeDirContents(dir string) error {
```
agis: Since this is only used in a single place, perhaps it'd be better to inline this in `Copy()`

cmd/mistry/transport.go
```diff
@@ -2,30 +2,63 @@ package main
...
-func (ts Scp) Copy(user, host, project, src, dst string) []string {
-	return []string{"scp", "-r", fmt.Sprintf("%s@%s:%s", user, host, src), dst}
+func (ts Scp) Copy(user, host, project, src, dst string, clearDst bool) (string, error) {
+	if clearDst {
+		err := removeDirContents(dst)
+		if err != nil {
+			return "", err
+		}
+	}
+	out, err := utils.RunCmd([]string{"scp", "-r", fmt.Sprintf("%s@%s:%s", user, host, src), dst})
```
agis: No need for these variables.

cmd/mistry/transport.go
```diff
@@ -2,30 +2,63 @@ package main
...
 	}
 	src = src[idx:]
+	cmd := []string{"rsync", "-rtlp"}
+	if clearDst {
+		cmd = append(cmd, "--delete")
+	}
+	cmd = append(cmd, fmt.Sprintf("%s@%s::%s/%s", user, host, module, src), dst)
 
-	return []string{"rsync", "-rtlp", fmt.Sprintf("%s@%s::%s/%s", user, host, module, src), dst}
+	out, err := utils.RunCmd(cmd)
```
agis: No need for these variables.

cmd/mistryd/mistryd_test.go
```diff
@@ -29,16 +29,20 @@ var (
 	params  = make(types.Params)
 
 	// mistry-cli args
-	host     string
-	port     string
-	username string
-	target   string
+	cliDefaultArgs CliCommonArgs
```
agis: I *think* you could drop `CliCommonArgs` and define the type right in here.

dtheodor: Yes that would have been ok. However I need the CliCommonArgs named struct as it is the input argument to `func cliBuildJobArgs(cliArgs CliCommonArgs, ...`

PR #65, opened 2018-04-30
Output the jobs logs at the API and CLI output

dtheodor:
Added the job logs as the `Log` attribute to the `BuildInfo`, so all API responses returning the BuildInfo  can provide job logs, in particular the `POST /jobs/` one. The CLI build invocation uses the response to output logs if the `--verbose` flag is set.

* Updated structs:
  * the Job already contained the BuildInfo (but serialized) as Job.Output. Removed that and added a proper struct Job.BuildInfo
  * move the Log to the BuildInfo, remove it from Job.Log
  * populate the BuildInfo with the log contents wherever is needed
  * update HTML template to use the updated structs
* move common code to read a job's BuildInfo JSON to a couple of new job functions. There was relevant code duplication when listing jobs, in the ExitCode function, and in tests
* update CLI to output the job logs

CLI output looks like this

screenshot from 2018-04-30 18-20-33

### DIFF
--- a/cmd/mistryd/job.go
+++ b/cmd/mistryd/job.go
@@ -7,8 +7,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"html/template"
 	"io"
+	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
@@ -56,10 +56,8 @@ type Job struct {
 
 	StartedAt time.Time
 
-	// webview-related
-	Output string
-	Log    template.HTML
-	State  string
+	BuildInfo *types.BuildInfo
+	State     string
 }
 
 // NewJobFromRequest returns a new Job from the JobRequest
@@ -119,13 +117,15 @@ func NewJob(project string, params types.Params, group string, cfg *Config) (*Jo
 	j.PendingBuildPath = filepath.Join(j.RootBuildPath, "pending", j.ID)
 	j.ReadyBuildPath = filepath.Join(j.RootBuildPath, "ready", j.ID)
 	j.ReadyDataPath = filepath.Join(j.ReadyBuildPath, DataDir)
-	j.BuildLogPath = filepath.Join(j.PendingBuildPath, BuildLogFname)
+	j.BuildLogPath = BuildLogPath(j.PendingBuildPath)
 	j.BuildInfoFilePath = filepath.Join(j.PendingBuildPath, BuildInfoFname)
 
 	j.Image = ImgCntPrefix + j.ID
 	j.Container = ImgCntPrefix + j.ID
 
 	j.StartedAt = time.Now()
+	j.BuildInfo = new(types.BuildInfo)
+	j.State = "pending"
 
 	return j, nil
 }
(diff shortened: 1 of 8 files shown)


cmd/mistryd/job.go
```diff
@@ -56,10 +56,8 @@ type Job struct {
 
 	StartedAt time.Time
 
-	// webview-related
-	Output string
-	Log    template.HTML
-	State  string
+	BuildInfo types.BuildInfo
```
agis: I think semantically it's more correct to have a pointer here.

cmd/mistryd/job.go
```diff
@@ -359,3 +353,55 @@ func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) (bool
 func (j *Job) RemoveBuildDir(fs filesystem.FileSystem, log *log.Logger) error {
 	return fs.Remove(j.ReadyBuildPath)
 }
+
+// ReadLogs returns the job's logs
+func (j *Job) ReadLogs() ([]byte, error) {
+	return ReadJobLogs(j.ReadyBuildPath)
```
agis: What if someone calls this when the job is pending?

cmd/mistryd/job.go
```diff
@@ -359,3 +353,55 @@ func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) (bool
 func (j *Job) RemoveBuildDir(fs filesystem.FileSystem, log *log.Logger) error {
 	return fs.Remove(j.ReadyBuildPath)
 }
+
+// ReadLogs returns the job's logs
+func (j *Job) ReadLogs() ([]byte, error) {
```
agis: I'd expect this method to just populate `j.BuildInfo.Log` instead of returning something.

cmd/mistryd/job.go
```diff
@@ -359,3 +353,55 @@ func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) (bool
...
+	return ReadJobLogs(j.ReadyBuildPath)
+}
+
+// GetBuildInfo returns the job's build info
+func (j *Job) GetBuildInfo(readLogs bool) (types.BuildInfo, error) {
```
agis: I think it's more intuitive to have do a `j.ReadBuildInfo()` that just populates `j.BuildInfo`.

cmd/mistryd/job.go
```diff
@@ -359,3 +353,55 @@ func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) (bool
...
+	return GetJobBuildInfo(j.ReadyBuildPath, readLogs)
+}
+
+// ReadyBuildLogPath returns the path of the job logs found at jobReadyPath
+func ReadyBuildLogPath(jobReadyPath string) string {
+	return filepath.Join(jobReadyPath, BuildLogFname)
```
agis: I think we can simplify this by checking both for `pending` and then `ready`.

cmd/mistryd/job.go
```diff
@@ -355,7 +351,44 @@ func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) (bool
...
+
+	log, err := ioutil.ReadFile(buildLogPath)
+	if err != nil {
+		return nil, err
+	}
+	return log, nil
+}
+
+// ReadJobBuildInfo returns the BuildInfo found at jobPath
+func ReadJobBuildInfo(jobPath string, readLogs bool) (*types.BuildInfo, error) {
```
agis: You could rename `jobPath` to the simpler `path`. Also `readLogs` to `logs`.

PR #72, opened 2018-05-04
fixed size worker pool to serve new job requests

dtheodor:
Based on https://gobyexample.com/worker-pools, with error and result handling improvements

* X amount of goroutines listen to a queue for new job requests
* HTTP new job requests place the work into the queue
* the result is communicated back to the HTTP request goroutine through
a per-request unique result channel

- [x] worker pool
- [x] configurable settings
- [x] tests
- [x] on exit wait for jobs to finish

### DIFF
--- a/cmd/mistryd/server.go
+++ b/cmd/mistryd/server.go
@@ -37,10 +37,11 @@ import (
 type Server struct {
 	Log *log.Logger
 
-	srv *http.Server
-	jq  *JobQueue
-	pq  *ProjectQueue
-	cfg *Config
+	srv        *http.Server
+	jq         *JobQueue
+	pq         *ProjectQueue
+	cfg        *Config
+	workerPool *WorkerPool
 
 	// web-view related
 
@@ -87,6 +88,7 @@ func NewServer(cfg *Config, logger *log.Logger) (*Server, error) {
 	s.pq = NewProjectQueue()
 	s.br = broker.NewBroker(s.Log)
 	s.tq = new(sync.Map)
+	s.workerPool = NewWorkerPool(s, cfg.Concurrency, cfg.Backlog, logger)
 	return s, nil
 }
 
@@ -118,20 +120,35 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if _, isAsync := r.URL.Query()["async"]; isAsync {
-		s.handleNewJobAsync(j, jr, w)
+	// send the work item to the worker pool
+	future, err := s.workerPool.SendWork(j)
+	if err != nil {
+		// the in-memory queue is overloaded, we have to wait for the workers to pick
+		// up new items.
+		// return a 503 to signal that the server is overloaded and for clients to try
+		// again later
+		// 503 is an appropriate status code to signal that the server is overloaded
+		// for all users, while 429 would have been used if we implemented user-specific
+		// throttling
+		s.Log.Print("Failed to send message to work queue")
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+
+	// if async, we're done, otherwise wait for the result in the result channel
+	_, async := r.URL.Query()["async"]
+	if async {
+		s.Log.Printf("Scheduled %s", j)
+		w.WriteHeader(http.StatusCreated)
 	} else {
-		s.handleNewJobSync(j, jr, w)
+		s.Log.Printf("Waiting for result of %s...", j)
+		s.writeWorkResult(j, future.Wait(), w)
 	}
 }
 
-// handleNewJobSync triggers the build synchronously, and writes the
-// build result JSON to the response
-func (s *Server) handleNewJobSync(j *Job, jr types.JobRequest, w http.ResponseWriter) {
-	s.Log.Printf("Building %s...", j)
-	buildInfo, err := s.Work(context.Background(), j, jr)
-	if err != nil {
-		http.Error(w, fmt.Sprintf("Error building %s: %s", j, err),
+func (s *Server) writeWorkResult(j *Job, r WorkResult, w http.ResponseWriter) {
+	if r.Err != nil {
+		http.Error(w, fmt.Sprintf("Error building %s: %s", j, r.Err),
 			http.StatusInternalServerError)
 		return
 	}
@@ -139,7 +156,7 @@ func (s *Server) handleNewJobSync(j *Job, jr types.JobRequest, w http.ResponseWr
 	w.WriteHeader(http.StatusCreated)
 	w.Header().Set("Content-Type", "application/json")
 
-	resp, err := json.Marshal(buildInfo)
+	resp, err := json.Marshal(r.BuildInfo)
 	if err != nil {
 		s.Log.Print(err)
 	}
@@ -147,12 +164,6 @@ func (s *Server) handleNewJobSync(j *Job, jr types.JobRequest, w http.ResponseWr
 	if err != nil {
 		s.Log.Printf("Error writing response for %s: %s", j, err)
 	}
-}
-
-func (s *Server) handleNewJobAsync(j *Job, jr types.JobRequest, w http.ResponseWriter) {
-	s.Log.Printf("Scheduling %s...", j)
-	go s.Work(context.Background(), j, jr)
-	w.WriteHeader(http.StatusCreated)
 }
 
 // HandleIndex returns all available jobs.
--- a/cmd/mistryd/worker_pool.go
+++ b/cmd/mistryd/worker_pool.go
@@ -1,0 +1,113 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"sync"
+
+	"github.com/skroutz/mistry/pkg/types"
+)
+
+// WorkResult contains the result of a job, either a buildinfo or an error
+type WorkResult struct {
+	BuildInfo *types.BuildInfo
+	Err       error
+}
+
+// FutureWorkResult is a WorkResult that may not yet have become available and
+// can be Wait()'ed on
+type FutureWorkResult struct {
+	result <-chan WorkResult
+}
+
+// Wait waits for the result to become available and returns it
+func (f FutureWorkResult) Wait() WorkResult {
+	r, ok := <-f.result
+	if !ok {
+		// this should never happen, reading from the result channel is exclusive to
+		// this future
+		panic("Failed to read from result channel")
+	}
+	return r
+}
+
+// workItem contains a job and a channel to place the job result. struct
+// used in the internal work queue
+type workItem struct {
+	job    *Job
+	result chan<- WorkResult
+}
+
+// WorkerPool implements a fixed-size pool of worker goroutines that can be sent
+// build jobs and communicate their result
+type WorkerPool struct {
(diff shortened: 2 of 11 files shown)


cmd/mistryd/server.go
```diff
@@ -118,28 +126,82 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// send the work item to the worker pool
+	workItem := WorkItem{j, jr, make(chan WorkResult, 1)}
```
agis: does this have to be buffered?

cmd/mistryd/server.go
```diff
@@ -118,28 +126,82 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusCreated)
 	} else {
-		s.handleNewJobSync(j, jr, w)
+		s.Log.Printf("Waiting for result for %s...", j)
+		result, ok := <-workItem.ResultChan
+		if !ok {
+			s.Log.Print("Failed to read from result channel")
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		s.Log.Printf("Received result for %s, returning", j)
```
agis: I think we're too verbose here.

cmd/mistryd/server.go
```diff
@@ -118,28 +126,82 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
...
+			close(item.ResultChan)
+		default:
+			// we can't place the result to the channel, this is a bug
+			s.Log.Printf("Failed to write to the result channel")
+		}
+	}
+	s.Log.Printf("%s work items channel closed, returning...", logPrefix)
+}
+
+func (s *Server) writeJobResultResponse(j *Job, r WorkResult, w http.ResponseWriter) {
```
agis: Are you sure this plays correct? I'd expect it to receive a pointer to the ResponseWriter, not a copy.

agis: Same goes for WorkResult.

agis: Perhaps `writeWorkResult()`?

cmd/mistryd/server.go
```diff
@@ -402,10 +458,18 @@ func (s *Server) HandleServerPush(w http.ResponseWriter, r *http.Request) {
 // non-nil error.
 func (s *Server) ListenAndServe() error {
 	s.Log.Printf("Configuration: %#v", s.cfg)
+	s.setupWorkerPool()
 	go s.br.ListenForClients()
 	return s.srv.ListenAndServe()
 }
 
+func (s *Server) setupWorkerPool() {
```
agis: This is too small and I think it should be inlined to `ListenAndServe`. We're not going to use it anywhere else for sure.

cmd/mistryd/server.go
```diff
@@ -118,28 +126,82 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
...
+}
+
+// WorkItem contains a job and a channel to place the job result
+type WorkItem struct {
+	Job        *Job
+	JobRequest types.JobRequest
+	ResultChan chan WorkResult
+}
+
+func (s *Server) worker(id int) {
```
agis: Needs a more accurate name (work?) and documentation. Also, do we need to have an `id`?

cmd/mistryd/server.go
```diff
@@ -118,28 +126,82 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
...
+	logPrefix := fmt.Sprintf("Worker %d:", id)
```
agis: We tend to prefix things like this:

> [worker] foo

cmd/mistryd/server.go
```diff
@@ -118,28 +126,82 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
...
-func (s *Server) handleNewJobSync(j *Job, jr types.JobRequest, w http.ResponseWriter) {
-	s.Log.Printf("Building %s...", j)
-	buildInfo, err := s.Work(context.Background(), j, jr)
-	if err != nil {
-		http.Error(w, fmt.Sprintf("Error building %s: %s", j, err),
+// WorkResult contains the result of a job, either a buildinfo or an error
+type WorkResult struct {
+	BuildInfo *types.BuildInfo
+	Err       error
+}
```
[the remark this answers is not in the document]
dtheodor: BuildInfo.Err contains build errors. The outer error is about failing to retrieve or construct the BuildInfo.

This struct captures the result of `Work() (BuildInfo, error)` which needs to be forwarded through a channel, so I'm capturing the two output parameters into a struct to pass it as one unit through channels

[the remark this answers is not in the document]
agis: I see, fair enough.

cmd/mistryd/server.go
dtheodor: ResponseWriter is an interface. The WorkResult doesn't need to be mutated so we don't need to pass a pointer

agis: j isn't mutated either.

cmd/mistryd/server.go
dtheodor: yes, the workers need to put the result there and move on to the next work item instead of waiting for the reader of this channel to receive the result, which may happen immediately, later, or never

cmd/mistryd/worker_pool.go
```diff
@@ -0,0 +1,102 @@
...
+	logger.Printf("Set up %d workers", maxWorkers)
+	return p
+}
+
+// Stop stops the pool workers and closes the queue
+func (p *WorkerPool) Stop() {
+	close(p.queue)
+}
+
+// SendWork sends work to the pool, and receives a work result that can be waited on.
```
agis: > SendWork schedules work on p and returns a FutureWorkResult. The actual result can be obtained by FutureWorkResult.Wait(). [...]

cmd/mistryd/worker_pool.go
```diff
@@ -0,0 +1,102 @@
...
+// An error is returned if the backlog is full and cannot accept any new work items
+func (p *WorkerPool) SendWork(j *Job, jr types.JobRequest) (FutureWorkResult, error) {
```
agis: Can we somehow make this agnostic to `Job` and `types.JobRequest`? I think we can. Couldn't this accept a `WorkItem`? This would seem more intuitive and better since we get rid of the tight coupling with the server.

cmd/mistryd/worker_pool.go
```diff
@@ -0,0 +1,102 @@
...
+	resultQueue := make(chan WorkResult, 1)
+	wi := workItem{j, jr, resultQueue}
+	result := FutureWorkResult{resultQueue}
+
+	select {
+	case p.queue <- wi:
+		return result, nil
+	default:
+		return result, fmt.Errorf("queue is full")
```
agis: No need to use `Errorf` since you don't format anything. Use `errors.New`. I think `go vet` should fail here anyway.

cmd/mistryd/worker_pool.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+// poolWorker listens to the workQueue, runs Work() on any incoming work items, and
+// sends the result through the result queue
+func poolWorker(s *Server, id int, workQueue <-chan workItem) {
```
agis: - Maybe just `spawn` or `start` is a better name?
- Also, same goes here regarding to `Server`. I think this is unnecessary coupling that we could and should avoid.
- "workQueue" could be renamed to the simpler "queue". The kind of thing that's queued in there is apparent from the type.

cmd/mistryd/worker_pool.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Wait waits for the result to become available and returns it
+func (future FutureWorkResult) Wait() (WorkResult, error) {
+	result, ok := <-future.resultQueue
+	if !ok {
+		return WorkResult{}, fmt.Errorf("Failed to read from result channel")
+	}
+	return result, nil
+}
```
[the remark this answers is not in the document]
dtheodor: We have no guarantee that it will not fail so we should handle it. The only way it can fail right now is if a bug is introduced. I considered `panic()`ing but I thought it's too much.

Even if panic() is acceptable at this API level, it is not acceptable to crash the mistry service from it, so I would have to write a panic error handler that returns an internal server error

cmd/mistryd/server.go
```diff
@@ -118,28 +120,49 @@ func (s *Server) HandleNewJob(w http.ResponseWriter, r *http.Request) {
...
+		w.WriteHeader(http.StatusCreated)
 	} else {
-		s.handleNewJobSync(j, jr, w)
+		s.Log.Printf("Waiting for result for %s...", j)
+		result, err := future.Wait()
+		if err != nil {
+			s.Log.Print("Failed to read from result channel")
+			w.WriteHeader(http.StatusInternalServerError)
+		} else {
+			s.Log.Printf("Received result for %s, returning", j)
```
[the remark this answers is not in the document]
dtheodor: this is debug logging. i'm removing it. we should deal with our logging strategy soon

cmd/mistryd/worker_pool.go
```diff
@@ -0,0 +1,102 @@
...
+
+// workItem contains a job and a channel to place the job result. struct
+// used in the internal work queue
+type workItem struct {
+	job        *Job
+	jobRequest types.JobRequest
```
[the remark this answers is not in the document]
dtheodor: i removed the JobRequest, only passing Job now

cmd/mistryd/worker_pool.go
```diff
@@ -0,0 +1,102 @@
...
+// sends the result through the result queue
+func poolWorker(s *Server, id int, workQueue <-chan workItem) {
+	logPrefix := fmt.Sprintf("[worker %d]", id)
+	for item := range workQueue {
+		s.Log.Printf("%s received work item %#v", logPrefix, item)
+		buildInfo, err := s.Work(context.Background(), item.job, item.jobRequest)
+
+		select {
+		case item.result <- WorkResult{buildInfo, err}:
+			s.Log.Printf("%s wrote result to the result channel", logPrefix)
```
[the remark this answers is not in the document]
dtheodor: Thinking to leave these logs for now, we can remove them after running on production for a couple of days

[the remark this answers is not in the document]
agis: Fair enough.

cmd/mistryd/worker_pool.go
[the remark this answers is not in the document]
agis: great!

cmd/mistryd/worker_pool_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+func setupQueue(t *testing.T, workers, backlog int) (*WorkerPool, *Config) {
+	cfg := testcfg
+	cfg.Concurrency = workers
+	cfg.Backlog = backlog
+
+	s, err := NewServer(cfg, nil)
+	failIfError(err, t)
```
apostergiou: :+1: 

As a side note, I opened an issue https://github.com/skroutz/mistry/issues/74 to package and refactor the test helpers.

apostergiou:
LGTM :+1:

PR #73, opened 2018-05-07
Fix streaming

### LINKED ISSUE #71: web view doesn't stream logs properly

apostergiou:
**Fix a bug with web view log streaming**
The bug was based in a racy condition that could be triggered when a
tailer had been terminated for a job ID just after a new job with the
same ID has started streaming.

To solve this we refactor the concurrency model of the tailers by
removing the sync map and depending only on new and closing clients
signals sent to the broker.

Fixes #71

Some other improvements:

- Fix a bug with clients number being negative
- broker: Close client channels when done

### DIFF
--- a/cmd/mistryd/server.go
+++ b/cmd/mistryd/server.go
@@ -2,7 +2,6 @@
 package main
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -17,9 +16,6 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
-	"time"
-
-	"sync"
 
 	dockerTypes "github.com/docker/docker/api/types"
 	dockerFilters "github.com/docker/docker/api/types/filters"
@@ -27,7 +23,6 @@ import (
 	"github.com/rakyll/statik/fs"
 	_ "github.com/skroutz/mistry/cmd/mistryd/statik"
 	"github.com/skroutz/mistry/pkg/broker"
-	"github.com/skroutz/mistry/pkg/tailer"
 	"github.com/skroutz/mistry/pkg/types"
 )
 
@@ -45,10 +40,6 @@ type Server struct {
 
 	// web-view related
 
-	// Queue used to track all open tailers by their id. Every tailer id
-	// matches a job id.
-	// The stored map type is [string]bool.
-	tq *sync.Map
 	br *broker.Broker
 	fs http.FileSystem
 }
@@ -87,7 +78,6 @@ func NewServer(cfg *Config, logger *log.Logger) (*Server, error) {
 	s.jq = NewJobQueue()
 	s.pq = NewProjectQueue()
 	s.br = broker.NewBroker(s.Log)
-	s.tq = new(sync.Map)
 	s.workerPool = NewWorkerPool(s, cfg.Concurrency, cfg.Backlog, logger)
 	return s, nil
 }
@@ -252,67 +242,6 @@ func (s *Server) HandleShowJob(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if state == "pending" {
-		// For each job id there is only one tailer responsible for
-		// emitting the read bytes to the s.br.Notifier channel.
-		_, ok := s.tq.Load(id)
-		if !ok {
-			// Create a channel to communicate the closure of all connections
-			// for the job id to the spawned tailer goroutine.
-			_, ok := s.br.CloseClientC[id]
-			if !ok {
-				s.br.CloseClientC[id] = make(chan struct{})
-			}
-
-			tl, err := tailer.New(BuildLogPath(jPath))
-			if err != nil {
-				s.Log.Print(err)
-				w.WriteHeader(http.StatusInternalServerError)
-				return
-			}
-
-			// Mark the id to the tailers' queue to identify that a
-			// tail reader has been spawned.
-			s.tq.Store(id, true)
-
-			go func() {
-				s.Log.Printf("[tailer] Starting for %s", id)
-
-				scanner := bufio.NewScanner(tl)
-				for scanner.Scan() {
-					s.br.Notifier <- &broker.Event{Msg: []byte(scanner.Text()), ID: id}
-				}
-			}()
-
-			go func() {
-				tick := time.NewTicker(time.Second * 3)
-				defer tick.Stop()
-
-			TAIL_CLOSE_LOOP:
-				for {
-					select {
-					case <-tick.C:
-						state, err := GetState(s.cfg.BuildPath, project, id)
-						if err != nil {
-							s.Log.Print(err)
-						}
-						if state == "ready" {
-							break TAIL_CLOSE_LOOP
-						}
-					case <-s.br.CloseClientC[id]:
-						break TAIL_CLOSE_LOOP
-					}
-				}
-				s.Log.Printf("[tailer] Exiting for: %s", id)
-				s.tq.Delete(id)
-				err = tl.Close()
-				if err != nil {
-					s.Log.Print(err)
-				}
-			}()
-		}
-	}
-
 	f, err := s.fs.Open("/templates/show.html")
 	if err != nil {
 		s.Log.Print(err)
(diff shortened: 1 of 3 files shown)


pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
 		case sb := <-br.NewClients:
 			br.clients[sb] = true
 			cc, ok := br.clientsCount.Load(sb.ID)
-			if ok {
+			if ok && cc.(int) > 0 {
```
agis: We should check here that the conversion works successfully. If we don't do so, this may panic, which is definitely unwanted.

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
 				br.clientsCount.Store(sb.ID, cc.(int)+1)
 			} else {
 				br.clientsCount.Store(sb.ID, 1)
+				br.CloseClientC[sb.ID] = make(chan struct{})
+				tl, err := tailer.New(sb.Extra)
```
agis: Error check?

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
+				go func() {
+					br.Log.Printf("[tailer] Starting for %s", sb.ID)
+
+					scanner := bufio.NewScanner(tl)
+					for scanner.Scan() {
+						br.Notifier <- &Event{Msg: []byte(scanner.Text()), ID: sb.ID}
+					}
+				}()
+				go func() {
+					<-br.CloseClientC[sb.ID]
```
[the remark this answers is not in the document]
apostergiou: Sure it would, that way we would notify the Tailer explicitly when it needs to terminate. 

Regardless of how we terminate the Tailer though, either explicitly with a channel message or implicitly with closing the file in question, I think the complexity would remain the same as we need the goroutine in order not to block while listening to the `br.CloseClientC` channel.

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
+					br.Log.Printf("[tailer] Exiting for: %s", sb.ID)
+					err = tl.Close()
+					if err != nil {
+						br.Log.Print(err)
+					}
+				}()
 			}
 			br.Log.Printf("[broker] Client added. %d registered clients", len(br.clients))
 		case sb := <-br.ClosingClients:
```
agis: Rename this variable to "client".

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
+			close(sb.EventC)
 			delete(br.clients, sb)
 			cc, ok := br.clientsCount.Load(sb.ID)
-			if ok {
+			if ok && cc.(int) > 0 {
```
agis: Check if conversion succeeds, otherwise this may panic.

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
 				br.clientsCount.Store(sb.ID, cc.(int)-1)
 			} else {
 				br.clientsCount.Store(sb.ID, 0)
 			}
 			br.Log.Printf("[broker] Removed client. %d registered clients", len(br.clients))
```
agis: Before we merge this and *after* making sure the logic works as expected, we should remove all these debug logs ("added client, removed client, closed tailer, new tailer etc.")

We should only log errors.

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
 			cc, ok = br.clientsCount.Load(sb.ID)
-			if ok && cc.(int) == 0 {
+			if ok && cc.(int) <= 0 {
```
agis: Check error from conversion or this may panic.

Also, is the `<= 0` check needed?

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
+					br.Log.Printf("[tailer] Exiting for: %s", sb.ID)
+					err = tl.Close()
+					if err != nil {
+						br.Log.Print(err)
+					}
+				}()
 			}
 			br.Log.Printf("[broker] Client added. %d registered clients", len(br.clients))
 		case sb := <-br.ClosingClients:
+			close(sb.EventC)
```
agis: Rename "EventC" to something better like "Messages" or "Data" or "Logs"

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
+			if ok && cc.(int) > 0 {
 				br.clientsCount.Store(sb.ID, cc.(int)+1)
 			} else {
 				br.clientsCount.Store(sb.ID, 1)
+				br.CloseClientC[sb.ID] = make(chan struct{})
+				tl, err := tailer.New(sb.Extra)
+				go func() {
+					br.Log.Printf("[tailer] Starting for %s", sb.ID)
+
+					scanner := bufio.NewScanner(tl)
```
agis: Since this variable is only used in the subsequent line, it could be renamed to the shorter "s".

pkg/broker/broker.go
```diff
@@ -42,6 +45,9 @@ type Client struct {
 	// Each connection has an id that corresponds to the Event ID it is
 	// interested in receiving messages about.
 	ID string
+
+	// Extra contains any extra misc information about the connection.
```
agis: Like what? Can we provide an example?

apostergiou: `Extra` may include information such as the path of a file to be read. Added an example.

cmd/mistryd/server.go
```diff
@@ -47,7 +42,6 @@ type Server struct {
 	// Queue used to track all open tailers by their id. Every tailer id
 	// matches a job id.
 	// The stored map type is [string]bool.
-	tq *sync.Map
```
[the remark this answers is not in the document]
apostergiou: true, done

cmd/mistryd/server.go
```diff
@@ -379,11 +311,13 @@ func (s *Server) HandleServerPush(w http.ResponseWriter, r *http.Request) {
...
 
-	client := &broker.Client{ID: id, EventC: make(chan []byte)}
+	jPath := filepath.Join(s.cfg.BuildPath, project, state, id)
+	buildLogPath := filepath.Join(jPath, BuildLogFname)
+	client := &broker.Client{ID: id, EventC: make(chan []byte), Extra: buildLogPath}
 	s.br.NewClients <- client
 
 	defer func() {
-		s.br.ClosingClients <- client
+		s.Log.Printf("Finished server push event at %s", r.URL.Path)
```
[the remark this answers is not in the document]
apostergiou: Yep, I used it to get some more information while debugging the handle closing but it's not needed in production.

pkg/broker/broker.go
```diff
@@ -76,23 +82,42 @@ func (br *Broker) ListenForClients() {
...
 			br.Log.Printf("[broker] Client added. %d registered clients", len(br.clients))
 		case sb := <-br.ClosingClients:
+			close(sb.EventC)
 			delete(br.clients, sb)
 			cc, ok := br.clientsCount.Load(sb.ID)
-			if ok {
+			if ok && cc.(int) > 0 {
 				br.clientsCount.Store(sb.ID, cc.(int)-1)
 			} else {
 				br.clientsCount.Store(sb.ID, 0)
```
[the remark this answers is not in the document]
apostergiou: Good catch, fixed.

pkg/broker/broker.go
apostergiou: Thanks for clarifying, fixed.

pkg/broker/broker.go
apostergiou: Added

pkg/broker/broker.go
apostergiou: done, it was about time we do this :smile:

pkg/broker/broker.go
apostergiou: fixed

pkg/broker/broker.go
apostergiou: sure

pkg/broker/broker.go
apostergiou: Added check for the type assertion. The `<= 0` is not needed indeed, it was a somewhat ugly patch for a past bug.

pkg/broker/broker.go
apostergiou: done

pkg/broker/broker.go
apostergiou: fixed

PR #77, opened 2018-05-09
Build result missing fix

### LINKED ISSUE #75: Build results are getting deleted when docker image build fails
Probably this can be resolved if we copy the pending builds to the ready path every time, no matter the type of failure.

apostergiou:
- [x] Introduce constructor for the BuildInfo
This constructor allows us to set the default ExitCode to 1 (overriding
the go's default 0).
- [x] Properly persist job results in case of failure
With this change, we copy the pending jobs directories to the ready path
whether they have failed or not. This makes sure we keep some
information (build_info.json and out.log files) about the failed jobs.
- [x] Refactor pending hanging jobs removal 
With this change we will handle hanging pending jobs by depending on
setted errors. This enables us to simplify the cleanup flow and drop the
`shouldCleanup` flag. As a side effect in order to be compatible with the current behavior we
take extra care of the LatestBuildPath which should point to the last
successful build.
- [x] Fix bug with overwriting build_info file
The new JSON that is going to overwrite the existing file may be
shorter in length. Thus, leaving some characters behind from the
previous write and making the JSON invalid.
We also remove the file descriptor since it not needed anywhere.

Fixes #75.

### DIFF
--- a/cmd/mistryd/end_to_end_test.go
+++ b/cmd/mistryd/end_to_end_test.go
@@ -283,7 +283,7 @@ func TestImageBuildFailure(t *testing.T) {
 	if err != nil {
 		t.Fatalf("%s", err)
 	}
-	assertEq(bi.ExitCode, -999, t)
+	assertEq(bi.ExitCode, types.ContainerFailureExitCode, t)
 }
 
 func TestLogs(t *testing.T) {
--- a/cmd/mistryd/end_to_end_test.go
+++ b/cmd/mistryd/end_to_end_test.go
@@ -258,6 +258,32 @@ func TestImageBuildFailure(t *testing.T) {
 	if !strings.Contains(cmderr, expErr) {
 		t.Fatalf("Expected '%s' to contain '%s'", cmderr, expErr)
 	}
+	j, err := NewJob("image-build-failure", types.Params{}, "", testcfg)
+	if err != nil {
+		t.Fatalf("failed to create job; %s", err)
+	}
+
+	errDockercmd := "apt-get install -y fofkoeakodksao"
+	log, err := ReadJobLogs(j.ReadyBuildPath)
+	if err != nil {
+		t.Fatalf("failed to read job log; %s", err)
+	}
+
+	if !strings.Contains(string(log), errDockercmd) {
+		t.Fatalf("Expected out log '%s' to contain '%s'", string(log), cmderr)
+	}
+
+	buildInfoPath := filepath.Join(j.ReadyBuildPath, BuildInfoFname)
+	bi := types.NewBuildInfo()
+	biBlob, err := ioutil.ReadFile(buildInfoPath)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+	err = json.Unmarshal(biBlob, bi)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+	assertEq(bi.ExitCode, -999, t)
 }
 
 func TestLogs(t *testing.T) {
--- a/cmd/mistryd/end_to_end_test.go
+++ b/cmd/mistryd/end_to_end_test.go
@@ -122,12 +122,12 @@ func TestAsyncSimpleBuild(t *testing.T) {
 		t.Fatalf("failed to find job build info at %s: %s", buildInfoPath, err)
 	}
 
-	bi := types.BuildInfo{}
+	bi := types.NewBuildInfo()
 	biBlob, err := ioutil.ReadFile(buildInfoPath)
 	if err != nil {
 		t.Fatalf("%s", err)
 	}
-	err = json.Unmarshal(biBlob, &bi)
+	err = json.Unmarshal(biBlob, bi)
 	if err != nil {
 		t.Fatalf("%s", err)
 	}
--- a/cmd/mistryd/job.go
+++ b/cmd/mistryd/job.go
@@ -416,7 +416,7 @@ func ReadJobLogs(jobPath string) ([]byte, error) {
 // ReadJobBuildInfo returns the BuildInfo found at jobPath
 func ReadJobBuildInfo(path string, logs bool) (*types.BuildInfo, error) {
 	buildInfoPath := filepath.Join(path, BuildInfoFname)
-	buildInfo := &types.BuildInfo{}
+	buildInfo := types.NewBuildInfo()
 
 	buildInfoBytes, err := ioutil.ReadFile(buildInfoPath)
 	if err != nil {
--- a/cmd/mistryd/job.go
+++ b/cmd/mistryd/job.go
@@ -354,10 +354,9 @@ func (j *Job) CloneSrcPath(log *log.Logger) string {
 	return cloneSrc
 }
 
-// BootstrapBuildDir creates all required build directories. Returns true if directories have
-// been partially or fully created and cleanup is required
-func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) (bool, error) {
-	shouldCleanup := false
+// BootstrapBuildDir creates all required build directories. Cleans the
+// pending directory if there were any errors.
+func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) error {
 	var err error
 
 	cloneSrc := j.CloneSrcPath(log)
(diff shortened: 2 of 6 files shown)


apostergiou:
It makes sense to populate the `BuildInfo` `Err` field in case of errors. But, to do this we would need to refactor the worker. I think we should track it in a separate issue.

@agis @dtheodor feedback is welcome.

pkg/types/build_info.go
```diff
@@ -38,6 +38,17 @@ type ErrImageBuild struct {
 	Err   error
 }
 
+func NewBuildInfo(params Params, path string, transportMethod TransportMethod, startedAt time.Time) *BuildInfo {
+	bi := new(BuildInfo)
+	bi.Params = params
+	bi.Path = path
+	bi.TransportMethod = transportMethod
+	bi.StartedAt = startedAt
+	bi.ExitCode = 1
```
dtheodor: I think we should use a special exit code value here to indicate the error condition that the container has not run and this exit code has not been returned by the container. So setting it to `1` doesn't allow us to determine whether this came from the container, or if this is a failure to even start the container

We already do this here
https://github.com/skroutz/mistry/blob/b02896475ff89e0e87f3798a0a477ce5325e7bd2/cmd/mistryd/worker.go#L282-L287

So I'm suggesting to initialize with `bi.ExitCode = -999`, and move the -999 value into a const

pkg/types/build_info.go
```diff
@@ -38,6 +38,17 @@ type ErrImageBuild struct {
 	Err   error
 }
 
+func NewBuildInfo(params Params, path string, transportMethod TransportMethod, startedAt time.Time) *BuildInfo {
```
agis: This shouldn't accept any parameters - the user should (if they want to) set individual fields after getting the result of `NewBuildInfo`.

It should just return a buildinfo with `ExitCode` and `StartedAt` initialized.

pkg/types/build_info.go
apostergiou: Sounds good.

pkg/types/build_info.go
apostergiou: done

cmd/mistryd/job.go
```diff
@@ -321,10 +321,9 @@ func (j *Job) CloneSrcPath(log *log.Logger) string {
 	return cloneSrc
 }
 
-// BootstrapBuildDir creates all required build directories. Returns true if directories have
-// been partially or fully created and cleanup is required
-func (j *Job) BootstrapBuildDir(fs filesystem.FileSystem, log *log.Logger) (bool, error) {
-	shouldCleanup := false
+// BootstrapBuildDir creates all required build directories. Cleanups the
```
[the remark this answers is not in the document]
apostergiou: fixed

cmd/mistryd/worker.go
```diff
@@ -98,10 +97,38 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
 	}
 
 	log.Printf("Creating new build directory...")
-	cleanupPending, err := j.BootstrapBuildDir(s.cfg.FileSystem, log)
+	err = j.BootstrapBuildDir(s.cfg.FileSystem, log)
+	if err != nil {
+		err = workErr("could not bootstrap build dir", err)
+	}
 
 	defer func() {
```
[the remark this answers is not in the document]
apostergiou: added

cmd/mistryd/worker.go
```diff
@@ -143,28 +170,12 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
-			}
-		}
-	}()
 	biJSON, err := json.Marshal(buildInfo)
 	if err != nil {
 		err = workErr("could not serialize build info", err)
 		return
 	}
-	_, err = infoFile.Write(biJSON)
+	err = ioutil.WriteFile(j.BuildInfoFilePath, biJSON, 0666)
```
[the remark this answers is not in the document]
apostergiou: There are constants for the file more https://golang.org/pkg/os/#FileMode. So, I think no change is needed.

cmd/mistryd/worker.go
```diff
@@ -188,60 +199,32 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
-		err = workErr("could not set the offset for the build_info file", err)
-		return
-	}
 	biJSON, err = json.Marshal(buildInfo)
 	if err != nil {
 		err = workErr("could not serialize build info", err)
 		return
 	}
-	_, err = infoFile.Write(biJSON)
+	err = ioutil.WriteFile(j.BuildInfoFilePath, biJSON, 0666)
```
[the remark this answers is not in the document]
apostergiou: same as before

cmd/mistryd/worker.go
```diff
@@ -98,10 +97,38 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
+			errstr := "could not move pending path"
+			if err == nil {
+				err = fmt.Errorf("%s; %s", errstr, rerr)
+			} else {
+				err = fmt.Errorf("%s; %s | %s", errstr, rerr, err)
+			}
+		}
+
+		if err == nil {
+			// LatestBuildPath always points to the last successful build.
```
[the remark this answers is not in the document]
apostergiou: reworded and moved the comment

cmd/mistryd/worker.go
```diff
@@ -98,10 +97,38 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
+					err = workErr("could not remove latest build link", err)
+					return
+				}
+			}
+			err = os.Symlink(j.ReadyBuildPath, j.LatestBuildPath)
+			if err != nil {
+				err = workErr("could not create latest build link", err)
+				return
+			}
+		} else {
```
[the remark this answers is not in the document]
apostergiou: added

cmd/mistryd/worker.go
```diff
@@ -98,10 +97,38 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
+		if rerr != nil {
+			errstr := "could not move pending path"
+			if err == nil {
+				err = fmt.Errorf("%s; %s", errstr, rerr)
+			} else {
+				err = fmt.Errorf("%s; %s | %s", errstr, rerr, err)
+			}
+		}
+
+		if err == nil {
```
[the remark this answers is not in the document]
apostergiou: added

agis:
> It makes sense to populate the BuildInfo Err field in case of errors. But, to do this we would need to refactor the worker. I think we should track it in a separate issue.

Yeah it makes sense to do so in another patch.

cmd/mistryd/worker.go
```diff
@@ -98,10 +97,40 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
+				err = fmt.Errorf("%s; %s", errstr, rerr)
+			} else {
+				err = fmt.Errorf("%s; %s | %s", errstr, rerr, err)
+			}
+		}
+
+		// When there are no errors the LatestBuildPath can be updated
+		// to point to the last build. Otherwise it remains unchanged.
+		if err == nil {
+			_, err = os.Lstat(j.LatestBuildPath)
```
dtheodor: What happens if the `err` from `os.Lstat` is not nil?

agis: Nice catch.

cmd/mistryd/end_to_end_test.go
```diff
@@ -223,6 +223,32 @@ func TestImageBuildFailure(t *testing.T) {
...
+	bi := types.NewBuildInfo()
+	biBlob, err := ioutil.ReadFile(buildInfoPath)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+	err = json.Unmarshal(biBlob, bi)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+	assertEq(bi.ExitCode, -999, t)
```
dtheodor: I'm replacing all references to `-999` with `types.ContainerFailureExitCode`

cmd/mistryd/worker.go
```diff
@@ -98,10 +97,40 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
+					return
+				}
+			}
+			err = os.Symlink(j.ReadyBuildPath, j.LatestBuildPath)
+			if err != nil {
+				err = workErr("could not create latest build link", err)
+			}
+			// An error at this point indicates that there is a hanging pending path.
+		} else {
 			derr := s.cfg.FileSystem.Remove(j.PendingBuildPath)
```
dtheodor: We don't need to `Remove(j.PendingBuildPath)` any more. We  always `os.Rename(j.PendingBuildPath, j.ReadyBuildPath)`. If that move fails, this means that we either have a bug or there is some system/permission error and manual intervention is needed by an admin to solve the issue. So I think we should never do  `Remove(j.PendingBuildPath)` here any more

agis: True.

PR #78, opened 2018-05-09
mistryd rebuild cli improvements

dtheodor:
* add `rebuild --verbose`, omit build logs when not true
* a bit more useful logs, with build times per project
* on failures, return non 0 exit code
* print the list of failed projects in the final output to avoid having to manually search the logs

### DIFF
--- a/cmd/mistryd/main.go
+++ b/cmd/mistryd/main.go
@@ -16,10 +16,8 @@ package main
 
 import (
 	"fmt"
-	"io"
 	"log"
 	"os"
-	"path/filepath"
 	"sync"
 
 	"github.com/skroutz/mistry/pkg/filesystem"
@@ -106,6 +104,9 @@ func main() {
 					Name:  "project, p",
 					Usage: "the project to build. Multiple projects can be specified. If not passed, all projects are built",
 				},
+				cli.BoolFlag{
+					Name: "verbose, v",
+				},
 			},
 			Action: func(c *cli.Context) error {
 				cfg, err := parseConfigFromCli(c.Parent())
@@ -113,17 +114,13 @@ func main() {
 					return err
 				}
 
-				logf, err := os.OpenFile(filepath.Join(cfg.BuildPath, "rebuild.log"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+				logger := log.New(os.Stdout, "", 0)
+				r, err := RebuildImages(cfg, logger, c.StringSlice("project"), c.Bool("fail-fast"), c.Bool("verbose"))
 				if err != nil {
-					return err
+					return fmt.Errorf("Image error, existing early, partial result: %s", r)
 				}
-				defer logf.Close()
-
-				out := io.MultiWriter(logf, os.Stdout)
-				logger := log.New(out, "", log.LstdFlags)
-				r, err := RebuildImages(cfg, logger, out, c.StringSlice("project"), c.Bool("fail-fast"))
-				if err != nil {
-					return fmt.Errorf("failed to build images with error: %s. Partial result: %s", err, r)
+				if len(r.failed) > 0 {
+					return fmt.Errorf("%s", r)
 				}
 				fmt.Printf("Finished. %s\n", r)
 				return nil
@@ -133,7 +130,8 @@ func main() {
 
 	err := app.Run(os.Args)
 	if err != nil {
-		log.Fatal(err)
+		fmt.Println(err)
+		os.Exit(1)
 	}
 }
 
--- a/cmd/mistryd/main.go
+++ b/cmd/mistryd/main.go
@@ -105,7 +105,8 @@ func main() {
 					Usage: "the project to build. Multiple projects can be specified. If not passed, all projects are built",
 				},
 				cli.BoolFlag{
-					Name: "verbose, v",
+					Name:  "verbose, v",
+					Usage: "print logs from docker build and run",
 				},
 			},
 			Action: func(c *cli.Context) error {
@@ -117,7 +118,7 @@ func main() {
 				logger := log.New(os.Stdout, "", 0)
 				r, err := RebuildImages(cfg, logger, c.StringSlice("project"), c.Bool("fail-fast"), c.Bool("verbose"))
 				if err != nil {
-					return fmt.Errorf("Image error, existing early, partial result: %s", r)
+					return err
 				}
 				if len(r.failed) > 0 {
 					return fmt.Errorf("%s", r)
--- a/cmd/mistryd/server.go
+++ b/cmd/mistryd/server.go
@@ -2,6 +2,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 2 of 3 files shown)


cmd/mistryd/server.go
```diff
@@ -379,34 +391,73 @@ func RebuildImages(cfg *Config, log *log.Logger, out io.Writer, projects []strin
...
-	r.pruneReport, err = client.ImagesPrune(ctx, dockerFilters.Args{})
+	r.prunedImages, r.prunedContainers, r.reclaimedSpace, err = dockerPruneUnused(ctx, client)
 	if err != nil {
 		return r, err
 	}
 	return r, nil
 }
 
+func dockerPruneUnused(ctx context.Context, c *docker.Client) (int, int, uint64, error) {
+	// TODO: we should look at filtering images and containers based on labels so that we
```
apostergiou: Labels seem nice. Consider opening a new issue for adding mistry related labels and remove the `TODO`.

cmd/mistryd/main.go
```diff
@@ -106,24 +104,23 @@ func main() {
...
 				if err != nil {
 					return err
 				}
 
-				logf, err := os.OpenFile(filepath.Join(cfg.BuildPath, "rebuild.log"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+				logger := log.New(os.Stdout, "", 0)
+				r, err := RebuildImages(cfg, logger, c.StringSlice("project"), c.Bool("fail-fast"), c.Bool("verbose"))
 				if err != nil {
-					return err
+					return fmt.Errorf("Image error, existing early, partial result: %s", r)
```
agis: Perhaps a simpler message:

> Error building image: %s

Also, how come we don't print `err` here?

cmd/mistryd/server.go
```diff
@@ -350,19 +352,29 @@ func (s *Server) ListenAndServe() error {
...
+	prunedImages     int
+	prunedContainers int
+	reclaimedSpace   uint64
 }
 
 func (r RebuildResult) String() string {
+	var failedNames string
+	if len(r.failed) > 0 {
+		failedNames = " Failed names: " + strings.Join(r.failed, ", ")
+	} else {
```
agis: Not needed.

cmd/mistryd/server.go
```diff
@@ -350,19 +352,29 @@ func (s *Server) ListenAndServe() error {
...
+		failedNames = ""
+	}
+
 	return fmt.Sprintf(
-		"Rebuilt: %d Failed: %d Pruned: %d Reclaimed: %d bytes",
-		r.successful, r.failed, len(r.pruneReport.ImagesDeleted), r.pruneReport.SpaceReclaimed)
+		"Rebuilt: %d Pruned images: %d Pruned containers: %d Reclaimed: %s Failed: %d%s",
```
agis: [nit] Add a separator (eg. comma) after each statistic for improved readability.

cmd/mistryd/server.go
```diff
@@ -379,34 +391,73 @@ func RebuildImages(cfg *Config, log *log.Logger, out io.Writer, projects []strin
...
 	}
-	r.pruneReport, err = client.ImagesPrune(ctx, dockerFilters.Args{})
+	r.prunedImages, r.prunedContainers, r.reclaimedSpace, err = dockerPruneUnused(ctx, client)
 	if err != nil {
 		return r, err
 	}
 	return r, nil
 }
 
+func dockerPruneUnused(ctx context.Context, c *docker.Client) (int, int, uint64, error) {
```
agis: Please document what the return values are and what the function does.

cmd/mistryd/server.go
```diff
@@ -350,19 +352,29 @@ func (s *Server) ListenAndServe() error {
...
-		"Rebuilt: %d Failed: %d Pruned: %d Reclaimed: %d bytes",
-		r.successful, r.failed, len(r.pruneReport.ImagesDeleted), r.pruneReport.SpaceReclaimed)
+		"Rebuilt: %d Pruned images: %d Pruned containers: %d Reclaimed: %s Failed: %d%s",
+		r.successful, r.prunedImages, r.prunedContainers, units.HumanSize(float64(r.reclaimedSpace)),
+		len(r.failed), failedNames)
 }
 
 // RebuildImages rebuilds images for all projects, and prunes any dangling images
-func RebuildImages(cfg *Config, log *log.Logger, out io.Writer, projects []string, failOnImageError bool) (RebuildResult, error) {
+func RebuildImages(cfg *Config, log *log.Logger, projects []string, failOnImageError, verbose bool) (RebuildResult, error) {
```
agis: [nit] Maybe `stopErr` would be a better name for that variable. The function does not fail, it returns immediately.

cmd/mistryd/server.go
```diff
@@ -379,34 +391,73 @@ func RebuildImages(cfg *Config, log *log.Logger, out io.Writer, projects []strin
...
-				log.Printf("Failed to build %s job %s with error: %s\n", project, j.ID, err)
+			var buildErr error
+			if verbose {
+				// pipe image build logs to the logger
+				pr, pw := io.Pipe()
+				buildResult := make(chan error)
+
+				go func() {
+					err := j.BuildImage(ctx, cfg.UID, client, pw, true, true)
+					pw.Close()
```
agis: What about the error?

cmd/mistryd/main.go
dtheodor: The `Action` function returns an error or nil, and the caller takes care of printing the error if one is returned

agis: I mean, we don't use the `err` variable at all.

cmd/mistryd/server.go
dtheodor: its returned as-is through the channel to the main thread, which takes care of dealing with it (fail early or append to list of failures)

agis: I mean, the error from `pw.Close()`

dtheodor: aha!

dtheodor: i looked into the implementation of PipeWriter.Close(), it *never* returns an error even if the function defines it as such. I will add a check here and just log the error if not nil and proceed. The error will always be nil at least with Go 1.10

cmd/mistryd/server.go
dtheodor: i will document the return values by just naming the return parameters

dtheodor: actually no, this creates unneccessary variables

dtheodor:
addressed comments

cmd/mistryd/server.go
agis: Cool.

cmd/mistryd/server.go
agis: That'd be nice. I don't think it does create unecessary variables.

agis:
@dtheodor We should make sure this PR enables T13350.

dtheodor:
This is for the server-side rebuild, 13350 is about better logs at the client side right?

agis:
Oops yeah I messed those up.

PR #81, opened 2018-05-11
add build --timeout

### LINKED ISSUE #67: Support setting timeouts
Right now, the client will block indefinitely if mistry is down (connect timeout) or a job takes forever to complete (read timeout).

However, there are cases when this is not acceptable. We should have the option to set a read timeout and a connect timeout on the client. Perhaps we only should expose this as a single timeout (how much the whole command should run for).

dtheodor:
Add a `mistry build --timeout 15s`. The client will exit with an error if the buidl takes longer than the timeout. The timeout string notation is defined at https://golang.org/pkg/time/#ParseDuration

connection timeouts (e.g. socket read() timeout and others) are already taken care of well enough by the `DefaultTransport` that is used in the http.Client by default

https://github.com/golang/go/blob/47be3d49c7d7ff77e675b0d0fb78c05fdb43dee2/src/net/http/transport.go#L39-L50

Relevant material https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/#clienttimeouts

Closes #67

### DIFF
--- a/cmd/mistry/main.go
+++ b/cmd/mistry/main.go
@@ -134,6 +134,7 @@ EXAMPLES:
 					Name:        "timeout",
 					Usage:       "time to wait for the build to finish, accepts values as defined at https://golang.org/pkg/time/#ParseDuration",
 					Destination: &timeout,
+					Value:       "60m",
 				},
 
 				// transport flags
--- a/cmd/mistry/main.go
+++ b/cmd/mistry/main.go
@@ -22,9 +22,11 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"os/user"
 	"strings"
+	"time"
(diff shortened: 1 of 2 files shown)


cmd/mistry/main.go
```diff
@@ -127,6 +130,11 @@ EXAMPLES:
 					Usage:       "rebuild the docker image",
 					Destination: &rebuild,
 				},
+				cli.StringFlag{
+					Name:        "timeout",
+					Usage:       "time to wait for the build to finish, accepts values as defined at https://golang.org/pkg/time/#ParseDuration",
```
agis: We could save the user from having to open a URL in order to use mistry. For this, we could introduce a reasonable default (60mins.) so that it's included in the help output.

> Maximum time to wait for a build.

PR #82, opened 2018-05-11
Rename unexpected containers whose name conflicts with mistry generated names

### LINKED ISSUE #20: Implement a strategy for stale containers
Right now, we remove containers in `StartContainer()` after they run successfully. However, it might happen, for example in case of a mistry failure while a container was running (eg. `SIGKILL`), that containers are left around with nobody to remove them.

We should decide on how to treat such cases.

dtheodor:
Closes #20

### DIFF
--- a/cmd/mistryd/job.go
+++ b/cmd/mistryd/job.go
@@ -3,7 +3,9 @@ package main
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
 	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -17,6 +19,7 @@ import (
 
 	dockertypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/mount"
 	docker "github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/jsonmessage"
@@ -184,6 +187,7 @@ func (j *Job) StartContainer(ctx context.Context, cfg *Config, c *docker.Client,
 
 	hostConfig := container.HostConfig{Mounts: mnts, AutoRemove: false, NetworkMode: "host"}
 
+	err := renameIfExists(ctx, c, j.Container)
 	res, err := c.ContainerCreate(ctx, &config, &hostConfig, nil, j.Container)
 	if err != nil {
 		return 0, err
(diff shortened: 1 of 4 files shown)


cmd/mistryd/job.go
```diff
@@ -233,6 +252,20 @@ func (j *Job) StartContainer(ctx context.Context, cfg *Config, c *docker.Client,
 	return result.State.ExitCode, nil
 }
 
+var conflictRe = regexp.MustCompile(`Conflict. The container name "\S+" is already in use by container "(\S+)".`)
```
agis: I'm concerned that this is too fragile and non-future-proof. Can't we just check if the container exists and if it does, delete it?

agis: And they don't expose the helper or something we could reuse?

dtheodor: I switched to a `renameIfExists` before creation. I'm renaming, I'm not sure we want to disrupt any existing containers. Any rogue containers will be garbage collected with our periodic prune

cmd/mistryd/job.go
```diff
@@ -233,6 +237,35 @@ func (j *Job) StartContainer(ctx context.Context, cfg *Config, c *docker.Client,
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func randomHexString() string {
+	buf := make([]byte, 16)
+	rand.Read(buf)
```
agis: Error check?

agis: Keep in mind that the default source you're using here...
> [...] produces a deterministic sequence of values each time a program is run.

(source: https://golang.org/pkg/math/rand)

dtheodor: rand.Read can never fail, not sure why they decided to return an error, maybe to keep the same API with other Read functions?

dtheodor: Ok, so if we restart the service, the same sequence of "random" values will be produced. I think this is OK. I could also use crypto/rand Read()

dtheodor: There's also this recipe `rand.Seed( time.Now().UTC().UnixNano())`

agis: This seems like a decent choice. You only need to do this once when the server boots though.

dtheodor: I'm initializing on mistry main() now, so any usage of rand called from the proper entrypoint (i.e. not from the tests) will be properly seeded

PR #85, opened 2018-05-14
output container stderr to the client build output

dtheodor:
* `mistry build` will output the container stderr if the container exit code is non zero
* `mistry build` will always output the URL at which the job logs are available
* The container stderr is seperately stored as `BuildInfo.ErrLog` (the `BuildInfo.Log` still contains all the logs) and returned with the POST new job response
* the client now logs its errors to stderr
* use Println in various client logs instead of Printf

Output of a job that fails:
screenshot from 2018-05-15 10-53-34

### DIFF
--- a/cmd/mistry/main.go
+++ b/cmd/mistry/main.go
@@ -234,7 +234,8 @@ EXAMPLES:
 					}
 				}
 
-				url := fmt.Sprintf("http://%s:%s/%s", host, port, JobsPath)
+				baseURL := fmt.Sprintf("http://%s:%s", host, port)
+				url := baseURL + "/" + JobsPath
 				if noWait {
 					url += "?async"
 				}
@@ -271,6 +272,10 @@ EXAMPLES:
 					return err
 				}
 
+				if !jsonResult {
+					fmt.Println("Logs can be found at", baseURL+"/"+bi.URL)
+				}
+
 				if verbose {
 					fmt.Printf(
 						"\nResult:\nStarted at: %s ExitCode: %v Params: %s Cached: %v Coalesced: %v\n\nLogs:\n%s\n",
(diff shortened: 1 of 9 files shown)


cmd/mistry/main.go
```diff
@@ -214,7 +214,8 @@ EXAMPLES:
 					}
 				}
 
-				url := fmt.Sprintf("http://%s:%s/%s", host, port, JobsPath)
+				baseUrl := fmt.Sprintf("http://%s:%s", host, port)
+				url := baseUrl + "/" + JobsPath
```
agis: Use https://golang.org/pkg/path/#Join

cmd/mistry/main.go
```diff
@@ -248,30 +249,37 @@ EXAMPLES:
...
 				}
 
 				if jsonResult {
-					fmt.Printf("%s", body)
+					fmt.Printf("%s\n", body)
 				}
 
 				if bi.ExitCode != 0 {
+					if bi.ErrLog != "" {
+						fmt.Fprintln(os.Stderr, "Container error logs:\n", bi.ErrLog)
```
agis: Maybe we should print this line even if `bi.ErrLog` is empty, so that it's apparent that there were actually no error logs?

As it is now, one might wonder why container error logs were not displayed.

cmd/mistry/main.go
```diff
@@ -248,30 +249,37 @@ EXAMPLES:
...
+					}
 					return fmt.Errorf("Build failed with exit code %d", bi.ExitCode)
 				}
 
 				if verbose {
-					fmt.Printf("Copying artifacts to %s", target)
+					fmt.Println("Copying artifacts to", target)
```
agis: Append a "..." here so that it's apparent that the operation might take some time.

cmd/mistry/main.go
```diff
@@ -248,30 +249,37 @@ EXAMPLES:
...
 				}
 				out, err := ts.Copy(transportUser, host, project, bi.Path+"/*", target, clearTarget)
 				fmt.Println(out)
 				if err != nil {
 					return err
 				}
 				if verbose {
-					fmt.Printf("Artifacts copied to %s successfully\n", target)
+					fmt.Println("Artifacts copied to", target, "successfully")
```
agis: "successfully" is redundant here imho.

pkg/types/build_info.go
```diff
@@ -32,7 +32,13 @@ type BuildInfo struct {
 
 	StartedAt time.Time
 
+	// Contains the stdout and stderr as output by the container
 	Log string
+	// Contains the stderr output by the container
```
agis: Insert a newline above.

agis:
Please also update the changelog.

cmd/mistry/main.go
dtheodor: doesn't work here

https://stackoverflow.com/questions/34668012/combine-url-paths-with-path-join

PR #92, opened 2018-06-05
work: Ensure that docker Client is closed

apostergiou:
This solves a bug where there are created too many file descriptors
under /proc/pid/fd with most of them being sockets. These file
descriptors are never freed so the service eventually stops responding
to new requests.

### DIFF
--- a/cmd/mistryd/worker.go
+++ b/cmd/mistryd/worker.go
@@ -182,6 +182,17 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo, 
 		err = workErr("could not create docker client", err)
 		return
 	}
+	defer func() {
+		derr := client.Close()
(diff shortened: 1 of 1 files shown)


cmd/mistryd/worker.go
```diff
@@ -191,12 +191,17 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
 		err = workErr("could not start docker container", err)
 		return
 	}
 
+	err = client.Close()
+	if err != nil {
+		err = workErr("could not close docker client", err)
+		return
+	}
+
```
agis: Maybe we need to defer this? As it is now, if `BuildImage` or `StartContainer` errors, it will never run.

apostergiou: True, I handled it with a defer function.

PR #97, opened 2018-08-01
Add support for opaque build params

0xdiba:
Any dynamic build params prepended with a `_` are not taken into account when calculating the job's ID.

### DIFF
--- a/cmd/mistry/main.go
+++ b/cmd/mistry/main.go
@@ -74,11 +74,12 @@ JOB PARAMETERS:
 EXAMPLES:
 	1. Schedule a job with a group and some parameters and put artifacts under
 		/tmp/yarn using rsync. Prefixing a file name with @ will cause the contents
-		of yarn.lock to be sent as parameters.
+		of yarn.lock to be sent as parameters. Parameters prepended with '_' are opaque
+		and do not affect the build result.
 
 		$ {{.HelpName}} --host example.org --port 9090 --project yarn \
 			--group group_name --transport rsync --target /tmp/yarn \
-			-- --lockfile=@yarn.lock --foo=bar
+			-- --lockfile=@yarn.lock --foo=bar --_ignored=true
 
 	2. Schedule a build and exit early without waiting for the result by setting
 		the no-wait flag.
--- a/cmd/mistryd/job_test.go
+++ b/cmd/mistryd/job_test.go
@@ -65,4 +65,15 @@ func TestJobID(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	opqParams := params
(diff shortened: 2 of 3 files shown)


cmd/mistry/main.go
```diff
@@ -74,11 +74,12 @@ JOB PARAMETERS:
 EXAMPLES:
 	1. Schedule a job with a group and some parameters and put artifacts under
 		/tmp/yarn using rsync. Prefixing a file name with @ will cause the contents
-		of yarn.lock to be sent as parameters.
+		of yarn.lock to be sent as parameters. Any dynamic options prepended with a '_'
```
ctrochalakis: To be consistent, let's call them parameters (or params) instead of dynamic options (see: https://github.com/skroutz/mistry/wiki/Glossary).

Also, `... will be treated as opaque for the result cache, but will be available during the build.` The build cache is the `/data/cache/` directory which is persisted between builds of the same group. The JobID hashing logic is called the result cache.

0xdiba:
Regarding the core param parsing logic, we have to take into account mistry's file parsing logic.
With the [current logic](https://github.com/skroutz/mistry/blob/master/cmd/mistry/main.go#L228) we cannot have opaque file params.

If this is a valid use case we should consider updating the code to support that ( could be just as simple as `|| strings.HasPrefix(v, "@") {` or change the whole opaque params handling logic cc @dtheodor :) )

cmd/mistry/main.go
```diff
@@ -74,11 +74,12 @@ JOB PARAMETERS:
 EXAMPLES:
 	1. Schedule a job with a group and some parameters and put artifacts under
 		/tmp/yarn using rsync. Prefixing a file name with @ will cause the contents
-		of yarn.lock to be sent as parameters.
+		of yarn.lock to be sent as parameters. Any parameters prepended with a '_'
```
agis: Consider simplifying the sentence:

> [...] as parameters. Parameters prepended with '_' are opaque and do not affect the build result.

cmd/mistryd/job_test.go
```diff
@@ -65,4 +65,15 @@ func TestJobID(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	ignParams := params
+	ignParams["_production"] = "ignored"
```
agis: Better use a consistent glossary. Let's call those opqParams (for "opaque").

cmd/mistry/main.go
0xdiba: Done

cmd/mistryd/job_test.go
0xdiba: Done

PR #107, opened 2018-09-26
Various improvements

agis:


cmd/mistryd/worker.go
```diff
@@ -116,9 +124,12 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
 			}
 		}
 
-		// When there are no errors the LatestBuildPath can be updated
-		// to point to the last build. Otherwise it remains unchanged.
+		// if the build was successful, symlink it to the 'latest'
+		// path
 		if err == nil {
+			s.pq.Lock(j.Project)
```
dtheodor: why lock?

cmd/mistryd/worker.go
```diff
@@ -201,40 +240,40 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
...
 		return
 	}
 
-	// fill the buildInfo Log from the freshly written log
-	err = out.Sync()
+	j.BuildInfo.Log = string(finalLog)
+	j.BuildInfo.ErrLog = outErr.String()
+	j.BuildInfo.Duration = time.Now().Sub(start).Truncate(time.Millisecond)
+
+	biJSON, err = json.Marshal(j.BuildInfo)
```
dtheodor: By moving JSON marshaling after filling the `.Log` attribute, the JSON now also contains the logs. This was not the case before. I remember with did this with @apostergiou on purpose, but I have forgotten why.

dtheodor: I think because the JSON was serialized to disk in a file andwe did not want the logs to be contained there

dtheodor: https://github.com/skroutz/mistry/pull/65

cmd/mistryd/server.go
```diff
@@ -494,30 +494,51 @@ func PruneZombieBuilds(cfg *Config) error {
...
 	if err != nil {
 		return nil, err
 	}
+
 	projects := []string{}
+
 	for _, f := range folders {
-		if f.IsDir() {
-			projects = append(projects, f.Name())
+		if !f.IsDir() {
```
0xdiba: The logic in this loop could be extracted to a separate method, sth like `isValidProject`, and could serve as a "definition" of a valid project.

cmd/mistryd/worker.go
agis: See https://github.com/skroutz/mistry/pull/107/commits/028a7d7fab1f023fd13b1d0d47fcd513921abc5c

cmd/mistryd/worker.go
agis: Makes sense. I think implementing a custom marshaler is the way here, so that we are explicit and say "we do *not* want logs to be marshaled".

README.md
```diff
@@ -163,6 +162,20 @@ For a sample configuration file refer to [`config.sample.json`](cmd/mistryd/conf
 
 
 
+Development
+---------------------------------------------------
+
+To run the tests, the Docker daemon is assumed to be running.
```
0xdiba: Could add an installation link here ( https://docs.docker.com/install/ )

cmd/mistryd/worker.go
dtheodor: can you document the "why" in the code? e.g.  "atomically modify the latest link to avoid race condition by concurrent builds of the same project"

cmd/mistryd/server.go
agis: I think this would make more sense if it was used in other places too. As it is now, I think the added level of indirection doesn't worth it.

cmd/mistryd/worker.go
```diff
@@ -286,3 +297,24 @@ func workErr(s string, e error) error {
...
+	bi := *j.BuildInfo
+	bi.ContainerStdouterr = ""
+	bi.ContainerStderr = ""
+
+	out, err := json.Marshal(bi)
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(j.BuildInfoFilePath, out, 0666)
```
dtheodor: simplified as `return ioutil.WriteFile(j.BuildInfoFilePath, out, 0666)` ?

Issue #124, opened 2019-10-18
Directories are not being added to docker's build context


linosgian:
**Problem:**
From  docker's documentation:

>The docker build command builds Docker images from a Dockerfile and a “context”. A build’s context is the set of files located in the specified PATH or URL. The build process can refer to any of the files in the context. For example, your build can use a COPY instruction to reference a file in the context.

The important bit from above is that we can only refer to files that are in the build's "context".
As seen in [mistry's source code](https://github.com/skroutz/mistry/blob/2b6c133b41e710f54c9ef70bd4b37fba2c7c7a6e/cmd/mistryd/job.go#L98), mistryd's `HandleNewJob` is invoked for each a new `Job`. In there `utils.Tar` is called on `j.ProjectPath`.

What [`utils.Tar` ](https://github.com/skroutz/mistry/blob/2b6c133b41e710f54c9ef70bd4b37fba2c7c7a6e/pkg/utils/utils.go#L61) does is  to create a tar containing the build context that's in turn passed to [Job.BuildImage](https://github.com/skroutz/mistry/blob/2b6c133b41e710f54c9ef70bd4b37fba2c7c7a6e/cmd/mistryd/job.go#L146) that uses docker's golang package.

The issue lies in the following lines of `utils.Tar`:
```
	walkFn := func(path string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}
		if !info.Mode().IsRegular() {
			return nil
		}
```
`walkFn` walks through the project's root, in order to collect all files and folders that the project needs. [FileMode.IsRegular](https://golang.org/pkg/os/#FileMode.IsRegular) will return False, if the file at hand is anything other than a "regular" file, aka directory, symlink etc.

The result here is that when trying to copy a **directory** from the project's root into the container's image, the building fails since we don't put non-regular files in the build context. An example looks like this:
```
# lgian[~/mistry/projects/brakeman] → l
total 24K
drwxr-xr-x  3 lgian lgian 4.0K Oct 18 15:52 .
drwxr-xr-x  3 lgian lgian 4.0K Oct 17 13:10 ..
-rw-r--r--  1 lgian lgian 1.5K Oct 17 13:45 docker-entrypoint
-rw-r--r--  1 lgian lgian  417 Oct 18 11:59 Dockerfile
-rw-r--r--  1 lgian lgian  574 Oct 17 13:11 ssh_known_hosts
[...]

PR #125, opened 2019-10-21
Pass relative paths to Docker build context

linosgian:
Docker accepts a so called "context" that contains all files that a
build can refer to. For instance, when `docker build -t example .` is
ran, all files under the current directory, are being tar-ed and copied
to a temporary location as the build's context.

In our case, in order to build a mistry project, docker's golang package
is being used. Docker client's [`ImageBuild`](https://godoc.org/docker.io/go-docker#Client.ImageBuild), takes an io.Reader
containing that build context. Again, a tarball with all the necessary
files.

The tarball we were creating was adding only the file's name as the
desired "untar" path to the tarball header, therefore,
all files were being untar-ed under the root folder.

The issue described above, led to `COPY` instructions failing, when
folders were being copied (or files under a certain directory for that matter).[Issue-124](https://github.com/skroutz/mistry/issues/124)

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -74,6 +74,12 @@ func Tar(root string) ([]byte, error) {
 			return err
 		}
 
+		// Preserve directory structure when docker "untars" the build context
+		hdr.Name, err = filepath.Rel(root, path)
+		if err != nil {
+			return err
+		}
+
 		err = tw.WriteHeader(hdr)
 		if err != nil {
 			return err
--- a/cmd/mistryd/end_to_end_test.go
+++ b/cmd/mistryd/end_to_end_test.go
@@ -367,6 +367,19 @@ func TestExitCode(t *testing.T) {
 	assert(br.ExitCode, 77, t)
 }
 
+func TestCopyDir(t *testing.T) {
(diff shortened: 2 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -74,6 +75,9 @@ func Tar(root string) ([]byte, error) {
 			return err
 		}
 
+		// Make untar destination relative, instead of absolute
+		hdr.Name = strings.TrimPrefix(strings.Replace(path, root, "", -1), string(filepath.Separator))
```
0xdiba: https://golang.org/pkg/path/filepath/#Rel probably does what you need here.

pkg/utils/utils.go
```diff
@@ -75,8 +74,11 @@ func Tar(root string) ([]byte, error) {
 			return err
 		}
 
-		// Make untar destination relative, instead of absolute
-		hdr.Name = strings.TrimPrefix(strings.Replace(path, root, "", -1), string(filepath.Separator))
+		// Make untar destination relative
```
agis: It'd be helpful to explain why we do this (i.e. to preserve the structure of the directory inside the tar)

PR #127, opened 2019-12-02
Returns err instead of nil when os.Stat returns err

gosom:


agis:
Hello, thanks for the PR.

The fact that we don't return an error if the path doesn't exist, is on purpose, as stated in the `FileSystem` interface:

https://github.com/skroutz/mistry/blob/f48d93fb2ce30e3ab9af6989c65fa23360ff37ad/pkg/filesystem/filesystem.go#L18-L21

So for the server, a non-existent directory is effectively a no-op.

gosom:
@agis yeap didn't notice that.

PR #129, opened 2020-12-03
A couple of additions to help newcomers.

jfragoulis:
Small additions to the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,19 @@ For an example refer to the [sample config](cmd/mistryd/config.sample.json).
 Development
 ---------------------------------------------------
 
+Before anything, make sure you install the dependencies:
+```shell
+make deps
+```
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -41,3 +41,11 @@ clean:
...
 generate:
 	go generate ./...
+
+docker_build:
+	docker build -t mistryd:latest .
+
+docker_run:
+	docker run --rm -p 8462:8462 -v /tmp/mistry:/var/lib/mistry -ti mistryd:latest
+
+run_mistryd: mistryd docker_build docker_run
```
jfragoulis, line 51: I chose this approach (makefile) instead of a `bin/mistryd` shell script (which was my initial intuition) because I think that having a single entry point for building and running all components of the project is proper.

agis (approved):

PR #130, opened 2020-12-04
Prometheus integration

### LINKED ISSUE #108: Support submitting build metrics
...to a backend like graphite.

jfragoulis:
This commit introduces a new class named `MetricsRecorder` and integrates with prometheus. Prometheus will pull the metrics every X seconds.

We have implemented:

* Builds on host
* Builds in general with state (success, failure, coalesce)
* Cache utilization
* Incremental utilization
* Build times

Closes #108

### DIFF
--- a/cmd/mistryd/metrics/metrics.go
+++ b/cmd/mistryd/metrics/metrics.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -13,7 +14,13 @@ import (
 type Recorder struct {
 	Log *log.Logger
 
-	BuildsHosted *prometheus.GaugeVec
+	BuildsHosted                 *prometheus.GaugeVec
+	BuildsStarted                *prometheus.CounterVec
+	BuildsFinished               *prometheus.CounterVec
+	BuildsCoalesced              *prometheus.CounterVec
+	BuildsProcessedIncrementally *prometheus.CounterVec
+	BuildsSucceeded              *prometheus.HistogramVec
+	BuildsFailed                 *prometheus.HistogramVec
 }
 
 const namespace = "mistry"
@@ -28,6 +35,66 @@ func NewRecorder(logger *log.Logger) *Recorder {
 			Namespace: namespace,
 			Name:      "builds_hosted",
 			Help:      "The number of finished build hosted currently in the server",
+		},
+		[]string{"project"},
+	)
+
+	r.BuildsStarted = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "builds_started",
+			Help:      "The total number builds started by the server",
+		},
+		[]string{"project"},
+	)
+
+	r.BuildsFinished = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "builds_finished",
+			Help:      "The number of builds finished.",
+		},
+		[]string{"project"},
+	)
+
+	r.BuildsCoalesced = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "builds_coalesced",
+			Help:      "The number of builds that coalesced and were not processed",
+		},
+		[]string{"project"},
+	)
+
+	r.BuildsProcessedIncrementally = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "builds_processed_incrementally",
+			Help:      "The number builds processed incrementally by the server",
+		},
+		[]string{"project"},
+	)
+
+	// The buckets we create start at 2 minutes and we create 3 buckets of
+	// 2 minute intervals.
+	buildTimeBuckets := prometheus.LinearBuckets(120, 120, 3)
+
+	r.BuildsSucceeded = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "builds_succeeded_seconds",
+			Help:      "Build duration and count for successful results.",
+			Buckets:   buildTimeBuckets,
+		},
+		[]string{"project"},
+	)
+
+	r.BuildsFailed = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "builds_failed_seconds",
+			Help:      "Build duration and count for failed results.",
+			Buckets:   buildTimeBuckets,
 		},
 		[]string{"project"},
 	)
@@ -56,5 +123,38 @@ func (r *Recorder) RecordHostedBuilds(buildPath, projectsPath string) {
 
 		labels := prometheus.Labels{"project": project.Name()}
 		r.BuildsHosted.With(labels).Set(float64(len(builds)))
+	}
+}
+
+// RecordBuildStarted records a build started, independently from its outcome.
+func (r *Recorder) RecordBuildStarted(project string) {
+	r.BuildsStarted.With(prometheus.Labels{"project": project}).Inc()
+}
+
+// RecordBuildCoalesced records a project's build when coalesced.
+func (r *Recorder) RecordBuildCoalesced(project string) {
+	r.BuildsCoalesced.With(prometheus.Labels{"project": project}).Inc()
+}
+
+// RecordBuildFinished records a project's state, whether it was build incrementally
+// and tis duration.
+func (r *Recorder) RecordBuildFinished(
+	project string,
+	success bool,
+	incremental bool,
+	duration time.Duration,
+) {
+	labels := prometheus.Labels{"project": project}
+
+	r.BuildsFinished.With(labels).Inc()
+
+	if success {
+		if incremental {
+			r.BuildsProcessedIncrementally.With(labels).Inc()
+		}
+
+		r.BuildsSucceeded.With(labels).Observe(duration.Seconds())
+	} else {
+		r.BuildsFailed.With(labels).Observe(duration.Seconds())
 	}
 }
--- a/cmd/mistryd/metrics/metrics.go
+++ b/cmd/mistryd/metrics/metrics.go
@@ -21,6 +21,7 @@ type Recorder struct {
 	BuildsProcessedIncrementally *prometheus.CounterVec
 	BuildsSucceeded              *prometheus.HistogramVec
 	BuildsFailed                 *prometheus.HistogramVec
+	CacheUtilization             *prometheus.CounterVec
 }
 
 const namespace = "mistry"
@@ -99,6 +100,15 @@ func NewRecorder(logger *log.Logger) *Recorder {
 		[]string{"project"},
 	)
 
+	r.CacheUtilization = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "cache_utilization",
+			Help:      "Build result cache utilization",
+		},
+		[]string{"project"},
+	)
+
 	return r
 }
 
@@ -157,4 +167,9 @@ func (r *Recorder) RecordBuildFinished(
 	} else {
 		r.BuildsFailed.With(labels).Observe(duration.Seconds())
 	}
+}
+
+// RecordCacheUtilization records a project build's cache utilization.
+func (r *Recorder) RecordCacheUtilization(project string) {
+	r.CacheUtilization.With(prometheus.Labels{"project": project}).Inc()
 }
--- a/cmd/mistryd/metrics/metrics.go
+++ b/cmd/mistryd/metrics/metrics.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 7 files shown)


cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+
+		time.Sleep(5 * time.Minute)
+	}
+}
+
+// Read the number of builds for the project by counting the folders under
+// its directories.
+func (r *MetricsRecorder) RecordBuildMetrics(project string) {
+	buildDir := fmt.Sprintf("%s/%s/ready", r.cfg.BuildPath, project)
```
jfragoulis, line 50: Is the number of builds going to count the contents of the `ready` folder or should we add the `pending` folder as well?

agis: So we have the "total number of builds" (an ever-increasing counter, counting all builds, pending+ready).

Now there are some extra dimensions:

1. A build can be either in-progress (pending) or completed (ready)
2. A completed build can be either successful or failed
3. A build can be either incremental or non-incremental

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	MetricsRecorderInstance *MetricsRecorder
+	MetricsRecorderOnce sync.Once
+)
+
+// Singleton access
+func GetMetricsRecorder() *MetricsRecorder {
```
agis, line 19: Instead of this function, we can just use a package-level global variable:

```go
var (
  MetricsRecorder = NewMetricsRecorder()
)

// ...
```

**Update**: Even better would be to make this just a member of the `Server` struct and initialize it inside `NewServer()`.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
+package main
```
agis, line 1: I think this should go into its own package, named `metrics` or `metrics_recorder`, as opposed to package `main`. Then the API would look like this:

```go
metrics.New() // or metrics_recorder.New()
// etc.
```

Note that it doesn't have to be in `<root>/pkg`, it can be located at `<root>/cmd/mistryd/metrics/`.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+	BuildTimes *prometheus.HistogramVec
+}
+
+func NewMetricsRecorder() *MetricsRecorder {
+	r := new(MetricsRecorder)
+
+	r.BuildsOnHost = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "mistry_builds_on_host",
+			Help: "The number of builds currently on the host",
```
agis, line 42: I think this needs some clarification. How about:

```
r.BuildsHosted
mistry_builds_hosted
The number of finished build hosted currently in the server
```

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		[]string{"project"},
+	)
+
+	r.Builds = promauto.NewCounterVec(
```
agis, line 47: Ditto:

```
r.BuildsProcessed
mistry_builds_processed
The number of builds processed by the server
```

cmd/mistryd/worker.go
```diff
@@ -55,6 +55,9 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
 					}
 					j.BuildInfo.ExitCode = i
 					j.BuildInfo.Coalesced = true
+
+					GetMetricsRecorder().RecordBuild(j.Project, "coalescing")
```
agis, line 59: s/coalescing/coalesced

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+			r.RecordBuildOnHost(project.Name(), buildPath)
+		}
+
+		time.Sleep(5 * time.Minute)
+	}
+}
+
+// Read the number of builds for the project by counting the folders under
+// its directories.
+func (r *MetricsRecorder) RecordBuildOnHost(project string, buildPath string) {
```
agis, line 99: The arguments can be simplified to `project, buildPath string`.

Ditto for all other functions in this file.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+	labels := prometheus.Labels{"project": project}
+	r.BuildsOnHost.With(labels).Set(float64(len(builds)))
+}
+
+func (r *MetricsRecorder) RecordBuild(project string, state string) {
+	labels := prometheus.Labels{"project": project, "state": state}
+	r.Builds.With(labels).Inc()
+}
+
+func (r *MetricsRecorder) RecordCacheUtilization(project string, cache string) {
```
agis, line 111: It'd be safer and more intuitive for this to read: `RecordCacheUtilization(project string, hit bool)`

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+	r.CacheUtilization.With(labels).Inc()
+}
+
+func (r *MetricsRecorder) RecordIncrementalUtilization(project string) {
+	labels := prometheus.Labels{"project": project}
+	r.IncrementalUtilization.With(labels).Inc()
+}
+
+func (r *MetricsRecorder) RecordBuildTime(project string, duration time.Duration) {
+	labels := prometheus.Labels{"project": project}
```
agis, line 122: We also need the result of the build (success/failure) here, because they typically differ a lot in execution times depending on whether the build was successful or failed early for some reason.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+	labels := prometheus.Labels{"project": project, "state": state}
+	r.Builds.With(labels).Inc()
+}
+
+func (r *MetricsRecorder) RecordCacheUtilization(project string, cache string) {
+	labels := prometheus.Labels{"project": project, "cache": cache}
+	r.CacheUtilization.With(labels).Inc()
+}
+
+func (r *MetricsRecorder) RecordIncrementalUtilization(project string) {
```
agis, line 116: Why does this have to be a separate metric? Can't it be a just a boolean label in the `RecordBuild()` function?

cmd/mistryd/worker.go
```diff
@@ -78,19 +81,25 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
 			// Previous build failed, remove its build dir to
 			// restart it. We know it's not pointed to by a
 			// latest link since we only symlink successful builds
+			GetMetricsRecorder().RecordCacheUtilization(j.Project, "hit-failed")
```
agis, line 84: s/hit-failed/miss

agis (changes requested):
I believe there are some unchecked errors. Please use [errcheck](https://github.com/kisielk/errcheck) to find them and handle them properly.

Bonus points for also running [staticcheck](https://staticcheck.io/docs) (but limit both this and errcheck to the newly-added code only).

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+func GetMetricsRecorder() *MetricsRecorder {
+    MetricsRecorderOnce.Do(func() {
+        MetricsRecorderInstance = NewMetricsRecorder()
+    })
+
+    return MetricsRecorderInstance
+}
+
+type MetricsRecorder struct {
+	cfg *Config
```
agis, line 28: This is not used anywhere.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Read the number of builds for the project by counting the folders under
+// its directories.
+func (r *MetricsRecorder) RecordBuildOnHost(project string, buildPath string) {
+	buildDir := fmt.Sprintf("%s/%s/ready", buildPath, project)
+	builds, _ := ioutil.ReadDir(buildDir)
+	labels := prometheus.Labels{"project": project}
+	r.BuildsOnHost.With(labels).Set(float64(len(builds)))
+}
```
agis, line 104: Since this is only called by a single place, I believe it'd be simpler to inline this as opposed to making it a separate function.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+// collects the data and populates the metrics for prometheus.
+func (r *MetricsRecorder) RecordDiskRelatedMetrics(buildPath string, projectsPath string) {
+	for {
+		projects, _ := ioutil.ReadDir(projectsPath)
+
+		for _, project := range projects {
+			r.RecordBuildOnHost(project.Name(), buildPath)
+		}
+
+		time.Sleep(5 * time.Minute)
```
agis, line 93: This should be the responsibility of the caller, like spinning it as a goroutine already is. Something like this seems more intuitive to me:

```
go func() {
    metrics.RecordDiskRelatedMetrics(cfg.BuildPath, cfg.ProjectsPath)
    time.Sleep(5 * time.Minute)
}()
```

And I'd do this in `NewServer()` as said in a previous comment.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+			Help: "Build times",
+			Buckets: prometheus.LinearBuckets(100, 100, 3),
+		},
+		[]string{"project"},
+	)
+
+	return r
+}
+
+// A goroutine which runs in the background and continuously (every 5 minutes)
```
agis, line 83: [nit] this is not a goroutine.

cmd/mistryd/metrics_recorder.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		[]string{"project", "state"},
+	)
+
+	r.CacheUtilization = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "mistry_cache_utilization",
+			Help: "Cache utilization",
+		},
+		[]string{"project", "cache"},
```
agis, line 60: Better use "cache": hit|miss

agis (changes requested):

cmd/mistryd/metrics_recorder.go
jfragoulis: This is a singleton implementation, that is why there is a `Get` function here. The singleton approach is an alternative to having the recorder as part of the server. 

I have no problem of instantiating the recorder in the server. The metrics are only related to the server anyway.

agis: I know. My point is that we can avoid the singleton since by definition we'll access this instance only via the server anyway.

cmd/mistryd/metrics_recorder.go
jfragoulis: 1. It is more intuitive to have this all in a single function.
1. This is going to happen next to another function that reads the disk utilization. It will make more sense then to have the code separate to its own place.

jfragoulis:
Run linter:
```
➜  mistry git:(prometheus-integration) make lint
golint `go list ./... | grep -v /vendor/`
/home/fragoulis/go/src/github.com/skroutz/mistry/cmd/mistry/main.go:248:24: error strings should not be capitalized or end with punctuation or a newline
```
Run errcheck:
```
➜  mistry git:(prometheus-integration) errcheck ./...                    
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/end_to_end_test.go:199:20:	defer os.RemoveAll(target)
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/end_to_end_test.go:211:10:	f.Close()
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/job.go:161:23:	defer resp.Body.Close()
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/job.go:216:18:	defer logs.Close()
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/job.go:267:11:	rand.Read(buf)
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/job_test.go:43:11:	os.Remove(path) // in case there's a leftover from a previous run
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/job_test.go:48:17:	defer os.Remove(path)
../../../go/src/github.com/skroutz/mistry/cmd/mistryd/server.go:108:14:	r.Body.Close()
```

Both presented no warnings related the changed files.

jfragoulis:
`staticcheck` did not output any errors.

agis (commented):
There are various places where errors are not handled (you can use `errcheck` to handle them).

cmd/mistryd/worker.go
```diff
@@ -78,19 +81,25 @@ func (s *Server) Work(ctx context.Context, j *Job) (buildInfo *types.BuildInfo,
 			// Previous build failed, remove its build dir to
 			// restart it. We know it's not pointed to by a
 			// latest link since we only symlink successful builds
+			s.metrics.RecordCacheUtilization(j.Project, false)
```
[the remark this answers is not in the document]
jfragoulis: You are probably looking at a cached version.

cmd/mistryd/server.go
```diff
@@ -613,3 +621,15 @@ func (s *Server) getJobs() ([]Job, error) {
 
 	return jobs, nil
 }
+
+// SetupMetrics initializes metrics recording for prometheus.
+func (s *Server) SetupMetrics() {
+	s.metrics = metrics.NewRecorder()
```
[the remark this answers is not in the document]
jfragoulis: This is objective. Personally I disagree (it is obvious since I wrote this). I will do you a favor though.

jfragoulis:
> There are various places where errors are not handled (you can use `errcheck` to handle them).

There are no errors in the files I touched or code I added. Are you suggesting I fix the pre-existing errors in this PR?

cmd/mistryd/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+		[]string{"project", "cache"},
+	)
+
+	return r
+}
+
+// RecordDiskRelatedMetrics collects metrics from disk. Its purpose is to
+// run in the background every few minutes.
+func (r *Recorder) RecordDiskRelatedMetrics(buildPath, projectsPath string) {
+	projects, _ := ioutil.ReadDir(projectsPath)
```
agis, line 70: We should check the error here and at least log it and return immediately.

cmd/mistryd/metrics/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, project := range projects {
+		r.RecordBuildsHosted(project.Name(), buildPath)
+	}
+}
+
+// RecordBuildsHosted reads the number of builds for the project by counting
+// the folders under its directories.
+func (r *Recorder) RecordBuildsHosted(project, buildPath string) {
+	buildDir := fmt.Sprintf("%s/%s/ready", buildPath, project)
+	builds, _ := ioutil.ReadDir(buildDir)
```
agis, line 81: ditto: err check

cmd/mistryd/server.go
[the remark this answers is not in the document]
agis: No strong opinion on this so feel free to leave it as you wish.

cmd/mistryd/worker.go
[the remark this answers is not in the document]
agis: This is what I'm looking at, and I'm referring to line 101 (I still see it there btw):

g

jfragoulis:
A small final touch https://github.com/skroutz/mistry/pull/130/commits/c2b12ac251970bb9c1ae5408345c68e4c1b46540

We do not need a separate metric to count the failed/successfull builds. The Histogram holds a counter as well as the timings.

agis (approved):
👏🏽 🚀

NikosGkotsis (approved):

